PR #3, opened 2015-11-16
select the most recent api version instead of 4.1

timogoebel:
This PR changes the behavior of negotiate_revision to negotiate the most recent revision instead of basically enforcing 4.1 for anything newer.

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -417,7 +417,7 @@ module Fog
           # Negotiate the API revision
           if not revision
             rev = @connection.serviceContent.about.apiVersion
-            @connection.rev = [ rev, ENV['FOG_VSPHERE_REV'] || '4.1' ].min
+            @connection.rev = [ rev, ENV['FOG_VSPHERE_REV'] || '4.1' ].max
           end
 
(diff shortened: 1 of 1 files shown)


shlomizadok:
Tests failure seems unrelated?

timogoebel:
@shlomizadok : I think, tests fail only on ruby 1.8...

shlomizadok:
@timogoebel and it is because fog dropped support for 1.8 :)

plribeiro3000:
Yeah, we should not be testing against `1.8` anymore. :+1:

timogoebel:
@plribeiro3000 : What do you think about that change? Any chance that it can be merged? We are discussing, whether we need to add an option to Foreman to set the API revision or can rely on fog to "do the right thing".

plribeiro3000:
What do you mean rely on `fog` to do the right thing?

We are striving to always do the right `thing` but we are might not have a human resource to take care of some providers. =s

Is there something we can do to help you guys out?

timogoebel:
@plribeiro3000 : Thanks for merging, now fog "does the right thing". You can have a look at this PR https://github.com/theforeman/foreman/pull/2913 to see the discussion.

plribeiro3000:
@timogoebel I would say we need this change at `fog` as well since the provider was not entirely extracted yet. Could you do that?

PR #6, opened 2015-11-16
implement storage pods

timogoebel:
This PR adds storage pods (datastore clusters) to fog. You can create a new vm or clone an existing one.

### DIFF
--- a/lib/fog/vsphere/models/compute/storage_pods.rb
+++ b/lib/fog/vsphere/models/compute/storage_pods.rb
@@ -1,0 +1,22 @@
+require 'fog/core/collection'
+require 'fog/vsphere/models/compute/storage_pod'
+
+module Fog
+  module Compute
+    class Vsphere
+      class StoragePods < Fog::Collection
+        model Fog::Compute::Vsphere::StoragePod
+        attribute :datacenter
+
+        def all(filters = {})
+          load service.list_storage_pods(filters.merge(datacenter: datacenter))
+        end
+
+        def get(id)
+          requires :datacenter
+          new service.get_storage_pod(id, datacenter)
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -26,7 +26,13 @@ module Fog
                             get_raw_cluster(attributes[:cluster], attributes[:datacenter]).resourcePool
                           end
           vmFolder      = get_raw_vmfolder(attributes[:path], attributes[:datacenter])
-          vm            = vmFolder.CreateVM_Task(:config => vm_cfg, :pool => resource_pool).wait_for_completion
+          # if any volume has a storage_pod set, we deploy the vm on a storage pod instead of the defined datastores
+          pod = get_storage_pod(attributes)
+          if pod
+            vm = create_vm_on_storage_pod(pod, vm_cfg, vmFolder, resource_pool, attributes[:datacenter])
+          else
+            vm = create_vm_on_datastore(vm_cfg, vmFolder, resource_pool)
+          end
           vm.config.instanceUuid
         rescue => e
           raise e, "failed to create vm: #{e}"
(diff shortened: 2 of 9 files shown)


lib/fog/vsphere/compute.rb.orig
```diff
@@ -0,0 +1,473 @@
+require 'digest/sha2'
```
plribeiro3000: Ops. Remove this file please. =)

timogoebel: Done, you were about 40 seconds too fast. ;-)

plribeiro3000: :grimacing:

lib/fog/vsphere/models/compute/storage_pods.rb
```diff
@@ -0,0 +1,22 @@
+require 'fog/core/collection'
+require 'fog/vsphere/models/compute/storage_pod'
+
+module Fog
+  module Compute
+    class Vsphere
+      class StoragePods < Fog::Collection
+        model Fog::Compute::Vsphere::StoragePod
+        attr_accessor :datacenter
```
plribeiro3000: Why didn't you use the DSL `attribute` here?

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -26,7 +26,12 @@ def create_vm attributes = { }
                             get_raw_cluster(attributes[:cluster], attributes[:datacenter]).resourcePool
                           end
           vmFolder      = get_raw_vmfolder(attributes[:path], attributes[:datacenter])
-          vm            = vmFolder.CreateVM_Task(:config => vm_cfg, :pool => resource_pool).wait_for_completion
+          # if any volume has a storage_pod set, we deploy the vm on a storage pod instead of the defined datastores
+          if pool = get_storage_pod(attributes)
```
plribeiro3000: Could you do the assign and then check if the variable is `nil` in the conditional?

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -34,9 +39,47 @@ def create_vm attributes = { }
...
+            :storagePod => get_raw_storage_pod(storage_pod, datacenter),
+          )
+          storage_spec = RbVmomi::VIM::StoragePlacementSpec.new(
+            :type => 'create',
+            :folder => vmFolder,
+            :resourcePool => resource_pool,
+            :podSelectionSpec => pod_spec,
+            :configSpec => vm_cfg,
+          )
+          srm = @connection.serviceContent.storageResourceManager
```
plribeiro3000: I think you could use the getter here (`connection`). This would make it easier to change the internals without breaking the services if needed.

lib/fog/vsphere/models/compute/storage_pods.rb
timogoebel: No reason. I just copied it ;-) Let me fix this...

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -581,31 +587,57 @@ def vm_clone(options = {})
...
+              new_vm = begin
+                # Try and find the new VM (folder.find is quite efficient)
+                dest_folder.find(options['name'], RbVmomi::VIM::VirtualMachine) or raise Fog::Vsphere::Errors::NotFound
+              rescue Fog::Vsphere::Errors::NotFound
+                tries += 1
+                if tries <= 10 then
+                  sleep 15
+                  retry
+                end
+                nil
```
plribeiro3000: I think you could use Fog [`wait_for`](https://github.com/fog/fog-core/blob/master/lib/fog/core/wait_for.rb#L2) here.

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -569,9 +573,11 @@ def vm_clone(options = {})
                                                                       :pool => resource_pool,
                                                                       :diskMoveType => :moveChildMostDiskBacking)
           else
-            relocation_spec = RbVmomi::VIM.VirtualMachineRelocateSpec(:datastore => datastore_obj,
-                                                                      :pool => resource_pool,
+            relocation_spec = RbVmomi::VIM.VirtualMachineRelocateSpec(:pool => resource_pool,
                                                                       :transform => options['transform'] || 'sparse')
+            unless options.key?('storage_pod')
+              relocation_spec[:datastore] = datastore_obj unless datastore_obj.nil?
```
plribeiro3000: Could you check both conditionals in one `unless` only?
It would make it easier to follow whats happening.

lib/fog/vsphere/requests/compute/list_storage_pods.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        def raw_storage_pods(datacenter_name)
+          dc = find_raw_datacenter(datacenter_name)
+
+          @connection.serviceContent.viewManager.CreateContainerView({
+            :container  => dc,
+            :type       => ["StoragePod"],
+            :recursive  => true
+          }).view
+        end
```
plribeiro3000: I believe this method `raw_storage_pods` should be private since its only purpose is to be used within `list_storage_pods`. Am i wrong?

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -180,19 +223,22 @@ def controller_get_shared_from_options options
...
-        def create_disk disk, index = 0, operation = :add, controller_key = 1000
+        def create_disk disk, index = 0, operation = :add, controller_key = 1000, storage_pod_selected = nil
           if (index > 6) then
             _index = index + 1
           else
             _index = index
           end
+          # If we deploy the vm on a storage pool, datastore has to be an empty string
+          datastore = ''
+          datastore = "[#{disk.datastore}]" unless storage_pod_selected
```
plribeiro3000: I think this would read better if you did an `if ... else` like:

```ruby
if storage_pod_selected
  datastore = ''
else
  datastore = "[#{disk.datastore}]"
end
```

plribeiro3000:
@timogoebel Could you run `rubocop` agains your changes and see how it looks like?

timogoebel:
@plribeiro3000 : I added the first round of fixes. rubocop complains a lot. I think, we should fix this up in a separate PR.

plribeiro3000:
@timogoebel Sure. I just wanted to see an actual status on rubocop.

lib/fog/vsphere/requests/compute/get_storage_pod.rb
```diff
@@ -0,0 +1,30 @@
...
+            :container  => dc,
+            :type       => ["StoragePod"],
+            :recursive  => true
+          }).view.select{|pod| pod.name == name}.first
+        end
+      end
+
+      class Mock
+        def get_storage_pod(name, datacenter_name)
+        end
```
plribeiro3000: If you are not going to implement it i would say is better to not even define it since `fog` will define it on the fly with a custom exception telling it is not implemented.

plribeiro3000:
@timogoebel Thank you for your time! This is a great work.

I believe this provider is kind of old and was not being updated within `fog` so there is a lot of discrepancies and stuff that we need to update like the use of `camelCase` for variables names and other stuff. I don't think you need to change it in this PR but i think we should do it in other PR's if we can.

I'm totally ok to merge this. I will just wait for @jrgarcia opinion here.

timogoebel:
@plribeiro3000: Thanks! Let me fix the Mock thing first.

plribeiro3000:
Sure. ping here when you finish because github does not notify of push changes, only comments.

jrgarcia:
@timogoebel I'm ok with merging once complete.

@plribeiro3000 There definitely is quite a bit that needs to be updated. I wanted to make sure that this gem was working well enough separated before making any sweeping changes like the `camelCase` one you mentioned.

plribeiro3000:
@jrgarcia Make total sense!

timogoebel:
@plribeiro3000 : I think, we are ready to merge!

jrgarcia:
@timogoebel Did you add any tests for this functionality?

timogoebel:
@jrgarcia : No, I did not.

jrgarcia:
@timogoebel Ok, that's fine for now. In the future, it would be nice to have tests for a change like this. Let me check this with a real environment and I'll merge it in.

timogoebel:
@jrgarcia : Thanks, that's awesome. I hope, everything works for you. If I find the time, I will add the tests in another PR.

jrgarcia:
:+1:

PR #9, opened 2015-11-19
Add virtual machine summary

coliver:
This commit also adds VirtualMachineConfigSummary and VirtualMachineStorageSummary since they are linked off of VirtualMachineSummary

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ tmp
 *.a
 mkmf.log
 gemfiles/*.lock
+*.ruby-*
+*.idea
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -17,46 +17,47 @@ module Fog
         # In particular:
         #   GuestInfo: information about the guest operating system
         #   VirtualMachineConfigInfo: Access to the VMX file and configuration
-
-        attribute :name
-        # UUID may be the same from VM to VM if the user does not select (I copied it)
-        attribute :uuid
-        attribute :hostname
-        attribute :operatingsystem
-        attribute :public_ip_address, :aliases => 'ipaddress'
-        attribute :power_state,   :aliases => 'power'
-        attribute :tools_state,   :aliases => 'tools'
-        attribute :tools_version
-        attribute :mac_addresses, :aliases => 'macs'
-        attribute :hypervisor,    :aliases => 'host'
-        attribute :connection_state
-        attribute :mo_ref
-        attribute :path
-        attribute :memory_mb
-        attribute :cpus
-        attribute :corespersocket
-        attribute :interfaces
-        attribute :volumes
-        attribute :customvalues
-        attribute :overall_status, :aliases => 'status'
         attribute :cluster
+        attribute :connection_state
+        attribute :corespersocket
+        attribute :cpuHotAddEnabled
+        attribute :cpus
+        attribute :customvalues
         attribute :datacenter
-        attribute :resource_pool
-        attribute :instance_uuid # move this --> id
+        attribute :firmware
         attribute :guest_id
         attribute :hardware_version
-        attribute :scsi_controller # this is the first scsi controller. Right now no more of them can be used.
-        attribute :cpuHotAddEnabled
+        attribute :hostname
+        attribute :hypervisor,    :aliases => 'host'
+        attribute :instance_uuid # move this --> id
+        attribute :interfaces
+        attribute :mac_addresses, :aliases => 'macs'
+        attribute :memory_mb
         attribute :memoryHotAddEnabled
-        attribute :firmware
+        attribute :mo_ref
+        attribute :name
+        attribute :operatingsystem
+        attribute :overall_status, :aliases => 'status'
+        attribute :path
+        attribute :power_state,   :aliases => 'power'
+        attribute :public_ip_address, :aliases => 'ipaddress'
+        attribute :resource_pool
+        attribute :scsi_controller # this is the first scsi controller. Right now no more of them can be used.
+        attribute :summary # VirtualMachineSummary
+        attribute :tools_state,   :aliases => 'tools'
+        attribute :tools_version
+        # UUID may be the same from VM to VM if the user does not select (I copied it)
+        attribute :uuid
+        attribute :volumes
 
         def initialize(attributes={} )
           super defaults.merge(attributes)
-          self.instance_uuid ||= id # TODO: remvoe instance_uuid as it can be replaced with simple id
+          self.instance_uuid ||= id # TODO: remove instance_uuid as it can be replaced with simple id
           initialize_interfaces
           initialize_volumes
           initialize_customvalues
           initialize_scsi_controller
+          initialize_virtual_machine_summary
         end
 
         # Lazy Loaded Attributes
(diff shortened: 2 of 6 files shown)


lib/fog/vsphere/models/compute/server.rb
```diff
@@ -17,46 +17,47 @@ class Server < Fog::Compute::Server
...
+        attribute :power_state,   :aliases => 'power'
+        attribute :public_ip_address, :aliases => 'ipaddress'
+        attribute :resource_pool
+        attribute :scsi_controller # this is the first scsi controller. Right now no more of them can be used.
+        attribute :summary # VirtualMachineSummary
+        attribute :tools_state,   :aliases => 'tools'
+        attribute :tools_version
+        # UUID may be the same from VM to VM if the user does not select (I copied it)
+        attribute :uuid
+        attribute :volumes
```
plribeiro3000: Did you just sorted them? :heart:

jrgarcia: I will never complain about sorted attributes :+1:

.gitignore
```diff
@@ -20,4 +20,6 @@ tmp
 *.o
 *.a
 mkmf.log
-gemfiles/*.lock
\ No newline at end of file
+gemfiles/*.lock
+*.ruby-*
```
plribeiro3000: Why did you add this to `gitignore`?
I think they are good to have because all developers will be on the same version and thus discussion wont be affected by different behaviours through ruby versions.
And we can always trust `travis` to test all versions for us. =)

.gitignore
```diff
@@ -20,4 +20,6 @@ tmp
 *.o
 *.a
 mkmf.log
-gemfiles/*.lock
\ No newline at end of file
+gemfiles/*.lock
```
jrgarcia: :+1: Thanks for this! Can't believe I missed it...

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -319,6 +320,13 @@ def initialize_scsi_controller
             Fog::Compute::Vsphere::SCSIController.new(self.attributes[:scsi_controller])
           end
         end
+
+        def initialize_virtual_machine_summary
+          if attributes[:summary] && attributes[:summary].props
+            self.attributes[:summary] = Fog::Compute::Vsphere::VirtualMachineSummary.new(self.attributes[:summary].props)
```
plribeiro3000: Is this code initializing an association?

If so what do you think of using `fog` DSL for it?

See examples at this [test](https://github.com/fog/fog-core/blob/master/spec/fog_attribute_spec.rb#L52-L57).

lib/fog/vsphere/models/compute/virtual_machine_config_summary.rb
```diff
@@ -0,0 +1,35 @@
...
+        attribute :uuid
+        attribute :vmPathName
+
+        def initialize(attributes={})
+          super
+          #TODO: Other objects get initialized here
+          #TODO: For now they will be their rbvmomi counterparts
+          #TODO: self.ftInfo = FaultToleranceConfigInfo.new(attributes[:ftInfo].props) if attributes[:ftInfo] && attributes[:ftInfo].props
+          #TODO: self.product = VAppProductInfo.new(attributes[:product].props) if attributes[:product] && attributes[:product].props
+        end
```
plribeiro3000: Could you remove this code if it is not doing anything yet?
I know you will add something but i feel if we don't need it yet better not add it since it will add one extra loop when evaluating this class.
And there is also the fact that if you are thinking about associations `Fog` have a builtin DSL for it.

lib/fog/vsphere/models/compute/virtual_machine_summary.rb
```diff
@@ -0,0 +1,25 @@
...
+        attribute :runtime        #VirtualMachineRuntimeInfo
+        attribute :storage        #VirtualMachineStorageSummary
+        attribute :vm             #ManagedObjectReference to a VirtualMachine
+
+        def initialize(attributes={})
+          super
+          #TODO: Other objects should get initialized here
+          #TODO: For now they will be their rbvmomi equivalents by default
+          self.storage = VirtualMachineStorageSummary.new(attributes[:storage].props) if attributes[:storage] && attributes[:storage].props
+          self.config = VirtualMachineConfigSummary.new(attributes[:config].props) if attributes[:config] && attributes[:config].props
```
plribeiro3000: Same here for the association DSL usage.

.gitignore
coliver: I agree that having ruby-version and ruby-gemset in the project is good, but they were not in there already so I was sticking with what was in there.  Let me know if I should change this.

lib/fog/vsphere/models/compute/server.rb
coliver: So, instead of this method server should has_one virtual_machine_summary?

lib/fog/vsphere/models/compute/server.rb
coliver: Yes. I use a plugin for RubyMine (I recommend it) called, surprisingly, sort lines that makes it very easy. :)

lib/fog/vsphere/models/compute/virtual_machine_config_summary.rb
coliver: With the exception of the two objects I haven't defined yet, ftInfo and product, all of the other attributes are basic types (strings, bools, and ints).  I figured those was useful. If you want to wait until those are defined I can remove this.

.gitignore
coliver: Also, today I learned there is a ~/.gitignore.  I should have never pushed this file... :/ sorry.

plribeiro3000: Thats not a problem. =)
We do need a `.gitignore` file here anyway. :+1:

lib/fog/vsphere/models/compute/server.rb
plribeiro3000: Yep. This will get the attributes and save in another instance variable inside the object
and when calling the getter it will get those attributes and initialize an object for you out of the box.

lib/fog/vsphere/models/compute/virtual_machine_config_summary.rb
plribeiro3000: Got it, but i believe you can solve this code here using the DSL for associations as well.

lib/fog/vsphere/models/compute/server.rb
coliver: I tried playing with this last week but was unable to get it going.  Other than that test, is there an example in another fog module I could look at to get it working?  I scanned through aws and google but didn't see it being used.

plribeiro3000: Yeah, so that feature was added because of `fog-xenserver`. Take a look at its models [here](https://github.com/fog/fog-xenserver/tree/master/lib/fog/compute/xen_server/models). None of the other providers did started using this feature yet but once they do lots of code won't be needed anymore.

jrgarcia:
@coliver Sorry, I know this is a dated PR, but are you planning on making any of the changes suggested (or resolving the conflicts with the latest code)?

coliver:
I couldn't get that dsl working, so I ended up using https://github.com/rlane/rbvmomi to get what I needed for the project I was working on, so, no, sorry. I'll retract the pull request.

jrgarcia:
No problem! I'm sorry it didn't work out for you. Thanks for closing this!

Issue #10, opened 2015-11-25
issue in integrating in foreman


nagarjuna:
I am trying to integrate fog-vsphere gem in foreman(http://www.theforeman.org/) application. When i try to use, i am getting below error

config = {
  provider: 'vsphere',
  vsphere_ssl: true,
  vsphere_rev: '4.0', # also fails using 4.1
  vsphere_username: 'root',
  vsphere_password: 'somepassword',
  vsphere_server: 'vcenter01',
  vsphere_expected_pubkey_hash: 'hash'
}

connection = Fog::Compute.new(config)

TypeError: superclass mismatch for class Vsphere
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/fog-vsphere-0.1.1/lib/fog/vsphere/compute.rb:5:in `<module:Compute>'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/fog-vsphere-0.1.1/lib/fog/vsphere/compute.rb:4:in `<module:Fog>'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/fog-vsphere-0.1.1/lib/fog/vsphere/compute.rb:3:in `<top (required)>'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/polyglot-0.3.5/lib/polyglot.rb:65:in `require'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/activesupport-3.2.21/lib/active_support/dependencies.rb:251:in `block in require'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/activesupport-3.2.21/lib/active_support/dependencies.rb:236:in `load_dependency'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/activesupport-3.2.21/lib/active_support/dependencies.rb:251:in `require'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/fog-core-1.32.0/lib/fog/core/services_mixin.rb:28:in `require_service_provider_library'
	from /home/nagarjuna/.rvm/gems/ruby-1.9.3-p547/gems/fog-core-1.32.0/lib/fog/core/services_mixin.rb:14:in `new'
[...]

PR #11, opened 2015-11-25
Require fog-core 1.35.0

shlomizadok:
Currently we're requiring fog-core ~1.32 which may break as fog-core < 1.35 still has Vsphere class (see issue #10). Updating to fog-core ~> 1.35 solves it.

### DIFF
--- a/fog-vsphere.gemspec
+++ b/fog-vsphere.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 1.8.7'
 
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Actually, i don't think we should even have a version locked here. If all providers start doing a lock, fog will never install because of campatibilities.
`fog` should be the one to lock `fog-core` version. 
Could you remove the lockin plz?

Thanks!

shlomizadok:
@plribeiro3000 - makes sense. removed. Thanks

plribeiro3000:
:+1: 

Thank you!

PR #13, opened 2015-11-27
implement cdrom options

timogoebel:
This PR adds code to manage cdrom drives with fog-vsphere.

I could not get the upload iso function to work, not sure if it's broken in rbvmomi or my test infrastructure. The fog code should be fine, though.

### DIFF
--- a/lib/fog/vsphere/models/compute/cdrom.rb
+++ b/lib/fog/vsphere/models/compute/cdrom.rb
@@ -1,0 +1,65 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Cdrom < Fog::Model
+        identity :key
+
+        attribute :filename
+        attribute :name, :default => "CD-/DVD-ROM Drive"
+        attribute :controller_key
+        attribute :unit_number
+        attribute :start_connected
+        attribute :allow_guest_control
+        attribute :connected
+
+        has_one_identity :server, :servers, :aliases => :instance_uuid
+
+        def to_s
+          name
+        end
+
+        def destroy
+          requires :instance_uuid, :key, :unit_number
+
+          service.destroy_vm_cdrom(self)
+          true
+        end
+
+        def save
+          requires :instance_uuid
+
+          if unit_number.nil?
+            used_unit_numbers = server.cdroms.map { |cdrom| cdrom.unit_number }
+            max_unit_number = used_unit_numbers.max
+
+            if max_unit_number > server.cdroms.size
+              # If the max ID exceeds the number of cdroms, there must be a hole in the range. Find a hole and use it.
+              self.unit_number = max_unit_number.times.to_a.find { |i| used_unit_numbers.exclude?(i) }
+            else
+              self.unit_number = max_unit_number + 1
+            end
+          else
+            if server.cdroms.any? { |cdrom| cdrom.unit_number == self.unit_number && cdrom.id != self.id }
+              raise "A cdrom already exists with that unit_number, so we can't save the new cdrom"
+            end
+          end
+
+          data = service.add_vm_cdrom(self)
+
+          if data['task_state'] == 'success'
+            # We have to query vSphere to get the cdrom attributes since the task handle doesn't include that info.
+            created = server.cdroms.get(unit_number)
+
+            self.key = created.key
+            self.filename = created.filename
+            self.unit_number = created.unit_number
+
+            true
+          else
+            false
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/vsphere/models/compute/cdroms.rb
+++ b/lib/fog/vsphere/models/compute/cdroms.rb
@@ -1,0 +1,18 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Cdroms < Fog::Collection
+        attribute :instance_uuid, :alias => :server_id
+        model Fog::Compute::Vsphere::Cdrom
+
+        def all(filters = {})
+          load service.list_vm_cdroms(instance_uuid)
+        end
+
+        def get(unit_number)
+          all.find { |cdrom| cdrom.unit_number == unit_number }
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -237,6 +237,14 @@ module Fog
           end
         end
 
+        def cdroms(opts = {})
+          service.cdroms(:instance_uuid => self.id).all(opts)
+        end
+
+        def cdrom(key)
+          cdroms.get(key)
+        end
+
         def guest_processes(opts = {})
           fail 'VM tools must be running' unless tools_running?
           service.list_processes(self.id, opts)
--- a/lib/fog/vsphere/requests/compute/list_vm_cdroms.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_cdroms.rb
@@ -1,0 +1,30 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+        def list_vm_cdroms(vm_id)
+          get_vm_ref(vm_id).config.hardware.device.select { |hw| hw.class == RbVmomi::VIM::VirtualCdrom }.map do |cdrom|
+            {
+              :filename => (cdrom.backing.fileName rescue(nil)),
+              :name => cdrom.deviceInfo.label,
+              :key => cdrom.key,
+              :controller_key => cdrom.controllerKey,
+              :unit_number => cdrom.unitNumber,
+              :start_connected => cdrom.connectable.startConnected,
+              :allow_guest_control => cdrom.connectable.allowGuestControl,
+              :connected => cdrom.connectable.connected,
+              :instance_uuid => vm_id,
+            }
+          end
+        end
+      end
+      class Mock
+        def list_vm_cdroms(vm_id)
+          raise Fog::Compute::Vsphere::NotFound, 'VM not Found' unless self.data[:servers].key?(vm_id)
+          return [] unless self.data[:servers][vm_id].key?('cdroms')
+          self.data[:servers][vm_id]['cdroms'].map {|h| h.merge({:instance_uuid => vm_id}) }
+        end
+      end
+    end
+  end
+end
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -43,6 +43,8 @@ module Fog
       collection :customfields
       model :scsicontroller
       model :process
+      model :cdrom
+      collection :cdroms
 
       request_path 'fog/vsphere/requests/compute'
       request :current_time
@@ -77,11 +79,14 @@ module Fog
       request :list_vm_interfaces
       request :modify_vm_interface
       request :modify_vm_volume
+      request :modify_vm_cdrom
       request :list_vm_volumes
+      request :list_vm_cdroms
       request :get_virtual_machine
       request :vm_reconfig_hardware
       request :vm_reconfig_memory
       request :vm_reconfig_cpus
+      request :vm_reconfig_cdrom
       request :vm_config_vnc
       request :create_folder
       request :list_server_types
@@ -97,6 +102,7 @@ module Fog
       request :list_child_snapshots
       request :revert_to_snapshot
       request :list_processes
+      request :upload_iso
 
       module Shared
         attr_reader :vsphere_is_vcenter
@@ -240,6 +246,17 @@ module Fog
                       "name"    => "Network adapter 1",
                       "status"  => "ok",
                       "summary" => "VM Network",
+                     }],
+                 "cdroms" =>
+                    [{
+                      "name"                => "CD-/DVD-Drive 1",
+                      "filename"            => nil,
+                      "key"                 => 3000,
+                      "controller_key"      => 200,
+                      "unit_number"         => 0,
+                      "start_connected"     => false,
+                      "allow_guest_control" => true,
+                      "connected"           => false,
                      }],
                  "hypervisor"       => "gunab.puppetlabs.lan",
                  "guest_id"         => "rhel6_64Guest",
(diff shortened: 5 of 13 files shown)


lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,83 @@
...
+        attribute :server_id
+        attribute :controller_key
+        attribute :unit_number
+        attribute :start_connected
+        attribute :allow_guest_control
+        attribute :connected
+
+        def initialize(attributes={} )
+          # Assign server first to prevent race condition with persisted?
+          self.server_id = attributes.delete(:server_id)
```
plribeiro3000: Whats the issue here?

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,83 @@
...
+      class Cdrom < Fog::Model
+        attribute :filename
+        attribute :name
+        attribute :key
+        attribute :server_id
+        attribute :controller_key
+        attribute :unit_number
+        attribute :start_connected
+        attribute :allow_guest_control
+        attribute :connected
```
plribeiro3000: Does this model have a unique id? perhaps `unit_number`?

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,83 @@
...
+            true
+          else
+            false
+          end
+        end
+
+        def server
+          requires :server_id
+          service.servers.get(server_id)
+        end
```
plribeiro3000: I think you could use the DSL for associations here. lie [here](https://github.com/fog/fog-xenserver/blob/master/lib/fog/compute/xen_server/models/server.rb#L86).

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        private
+
+        def defaults
+          {
+            :name => "CDROM Drive",
```
plribeiro3000: You could use the DSL to define this like:

```ruby
attribute :name, default => "CDROM Drive"
```

lib/fog/vsphere/models/compute/cdroms.rb
```diff
@@ -0,0 +1,43 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Cdroms < Fog::Collection
+        autoload :Cdrom, File.expand_path('../cdrom', __FILE__)
```
plribeiro3000: I guess this `autoload` needs to be inside the `Vsphere` class since it is not inside the `Cdroms` one.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -218,6 +220,10 @@ def volumes
           attributes[:volumes] ||= id.nil? ? [] : service.volumes(:server_id => self.id)
         end
 
+        def cdroms
+          attributes[:cdroms] ||= id.nil? ? [] : service.cdroms(:server_id => self.id)
+        end
```
plribeiro3000: You can use the DSL association here as well.

lib/fog/vsphere/requests/compute/list_vm_cdroms.rb
```diff
@@ -0,0 +1,26 @@
...
+              :allow_guest_control => cdrom.connectable.allowGuestControl,
+              :connected => cdrom.connectable.connected,
+            }
+          end
+        end
+      end
+      class Mock
+        def list_vm_cdroms(vm_id)
+        end
+      end
```
plribeiro3000: If you are not going to implement it don't define it because fog will raise an exception out of the box saying it is not defined which is far better then doing nothing.

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: I think, I simply forgot to clean that up...

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: unit_number is just unique per controller, we could use key I think.
Are you thinking about ```identity :key```?

plribeiro3000: Yeap. Because `persisted?` uses it.

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: You mean I can replace the function by this?
```
has_one_identity      :server,     :servers
```

plribeiro3000: Yeap. Could you test it out?

I'm pretty sure it works just like xenserver and because of that it will work.

This method will define a getter that return an object instead of the id only.

lib/fog/vsphere/models/compute/server.rb
timogoebel: Not sure I can follow, do you mean something like this?
```
has_many_identities   :cdroms,     :cdroms
```

plribeiro3000: Exactly. =)
And it won't memoize so you won't need to check it when reloading.

lib/fog/vsphere/models/compute/cdroms.rb
timogoebel: Are you sure about this? It works fine like this and the volumes model does it like this as well...

timogoebel:
@plribeiro3000 : Thanks for your suggestions. Will keep them in mind for future PRs. :+1: 
I updated to code accordingly.

lib/fog/vsphere/models/compute/cdroms.rb
plribeiro3000: It might work but thats not the expected. `autoload` will define a constant on that scope.
So i would expect it to define a constant in the wrong place.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -37,6 +37,7 @@ class Server < Fog::Compute::Server
         attribute :corespersocket
         attribute :interfaces
         attribute :volumes
+        attribute :cdroms
```
plribeiro3000: You don't need this anymore since the `has_many_identities` will define it for you.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -50,11 +51,14 @@ class Server < Fog::Compute::Server
...
         attribute :firmware
 
+        has_many_identities :cdroms, :cdroms
+
         def initialize(attributes={} )
           super defaults.merge(attributes)
           self.instance_uuid ||= id # TODO: remvoe instance_uuid as it can be replaced with simple id
           initialize_interfaces
           initialize_volumes
+          initialize_cdroms
```
plribeiro3000: You don't need to initialize anymore because the association DSL will do that for you.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -272,7 +276,7 @@ def new?
 
         def reload
           # reload does not re-read assoiciated attributes, so we clear it manually
-          [:interfaces, :volumes].each do |attr|
+          [:interfaces, :volumes, :cdroms].each do |attr|
```
plribeiro3000: You don't need to check this anymore.

plribeiro3000:
@timogoebel Can you confirm it works using the association DSL?

And could you remove the `Mock` stuff since you are not implementing it?

timogoebel:
@plribeiro3000 :  I updated the code and added some tests. Mind to take a look?

plribeiro3000:
Sure, I'm taking a look right now sir!

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,66 @@
...
+            if server.cdroms.any? { |cdrom| cdrom.unit_number == self.unit_number && cdrom.id != self.id }
+              raise "A cdrom already exists with that unit_number, so we can't save the new cdrom"
+            end
+          end
+
+          data = service.add_vm_cdrom(self)
+
+          if data['task_state'] == 'success'
+            # We have to query vSphere to get the cdrom attributes since the task handle doesn't include that info.
+            created = server.cdroms.all.find { |cdrom| cdrom.unit_number == self.unit_number }
```
plribeiro3000: I guess you could use the `cdrom` collection `#get` here and avoid doing a search in ruby.

```ruby
created = server.cdroms.get(self.unit_number)
```

plribeiro3000:
@timogoebel Can you confirm the association DSL is working?

lib/fog/vsphere/models/compute/cdroms.rb
timogoebel: Had another look at this, all other models do the autoloading this way. Where would you define that, @plribeiro3000 ?

timogoebel:
@plribeiro3000 : I checked that, it's complaining about an undefined method 'server_id' when I drop the attribute statement..

plribeiro3000:
@timogoebel Even when you define the `alias`?

timogoebel:
@plribeiro3000 : Yes, even when I define the alias.

plribeiro3000:
But does the DSL works this way?
Are you able to get a server object back when calling `server`?

timogoebel:
@plribeiro3000: Another round of updates, refactored this a lot. Code looks much cleaner through heavy usage of the DSL. Love to hear your feedback.

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        attribute :filename
+        attribute :name, :default => "CD-/DVD-ROM Drive"
+        attribute :controller_key
+        attribute :unit_number
+        attribute :start_connected
+        attribute :allow_guest_control
+        attribute :connected
+
+        has_one_identity :server, :servers, :aliases => :instance_uuid
```
plribeiro3000: Does this work now?

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        def to_s
+          name
+        end
+
+        def destroy
+          requires :server_id, :key, :unit_number
```
plribeiro3000: I guess this will break since the attribute `server_id` was removed. Can you confirm?

lib/fog/vsphere/models/compute/cdrom.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          service.destroy_vm_cdrom(self)
+          true
+        end
+
+        def save
+          requires :server_id
```
plribeiro3000: Same thing here. Does this work after the removal of the `server_id` attribute?

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -237,6 +239,14 @@ def revert_snapshot(snapshot)
           end
         end
 
+        def cdroms(opts = {})
+          service.cdroms(:instance_uuid => self.id).all(opts)
+        end
```
plribeiro3000: I guess this method is overwriting the getter defined by the `has_many_identities` DSL.

Then it does not make much sense to have the DSL and this method i guess. What do you think?

plribeiro3000:
@timogoebel The code is getting better and better indeed. `fog` DSL is quite great and provide a lot of features that should help maintaining the code but since we don't have a good documentation about it, it takes time to swallow. Sorry about that.

Thank you for bearing with me here!

I just have a couple of questions regarding the `server_id` now and we are good to go. There is stuff to refactor on the request level as well but lets leave it for another PR. =)

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: Ack. Sorry, I missed that. :-)

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: Missed that, sorry...

lib/fog/vsphere/models/compute/server.rb
timogoebel: I think, we need to do define our own getter as we need to pass the instance_uuid. Or is there a way set the instance_uuid using the DSL?

lib/fog/vsphere/models/compute/cdrom.rb
timogoebel: Yes, this does actually work. :-)

plribeiro3000: Awesome. =)

timogoebel:
> The code is getting better and better indeed.
Glad to hear that. The DSL is causing me some severe headaches ;-)

> There is stuff to refactor on the request level as well but lets leave it for another PR. =)
I'd appreciate to hear your thoughts.

lib/fog/vsphere/models/compute/server.rb
plribeiro3000: Yeap. it does not work with custom parameters yet.
I guess it would be a great thing but lets not worry about it now. =)

So jus remove the DSL part. =)

timogoebel: Ok. Pushed the new code.

plribeiro3000:
@timogoebel LGTM!

Regarding the requests there is the parameter `as` on the attribute DSL. It will transform the hash generated by the object to a hash using the expected keys by the provider.

Buts lets take a look at this later.

This PR took a lot of your time already. =)

Thank you!

timogoebel:
@plribeiro3000 : Thanks! It was a pleasure working with you on this.

plribeiro3000:
@timogoebel It was my pleasure. Hope to see you more often around here. =)

timogoebel:
@plribeiro3000 : Do you have any idea when new releases of fog and fog-vsphere are published containing these changes?

plribeiro3000:
@timogoebel I don't have permissions to do so but i believe @jrgarcia or @geemus can help you out here. =)

jrgarcia:
I'll look at releasing a new version today.

plribeiro3000:
Awesome. Thx!

Em qui, 3 de dez de 2015 12:27, J.R. Garcia <notifications@github.com>
escreveu:

geemus:
Thanks!

timogoebel:
@geemus : Can you tell more about a new release for "fog"? Is there a release schedule or something?

geemus:
Sure. Theoretically it is the first Thursday of the month, I just get over-run and fall behind sometimes (like now), so we are about a week overdue. Hopefully I can remedy either today or tomorrow though and get us more-or-less back on track.

PR #14, opened 2015-11-30
add mocking and basic test for storage pods

timogoebel:
This PR adds some basic mocking and a basic test... Nothing fancy, but a start...

### DIFF
--- a/lib/fog/vsphere/requests/compute/get_storage_pod.rb
+++ b/lib/fog/vsphere/requests/compute/get_storage_pod.rb
@@ -23,6 +23,7 @@ module Fog
 
       class Mock
         def get_storage_pod(name, datacenter_name)
+          list_storage_pods({datacenter: datacenter_name}).select{|h| h[:name] == name }.first
         end
       end
     end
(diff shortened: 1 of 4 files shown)


lib/fog/vsphere/requests/compute/get_storage_pod.rb
```diff
@@ -23,6 +23,7 @@ def get_raw_storage_pod(name, datacenter_name)
 
       class Mock
         def get_storage_pod(name, datacenter_name)
+          list_storage_pods().select{|h| h[:name] == name }.first
```
plribeiro3000: I believe the mock version of the api should be more like the real one. Can't you do a search based on the `datacenter_name`?

plribeiro3000:
@timogoebel Looks awesome. Just a small question that i believe there is nothing to do. I just want to make sure i'm not missing it.

timogoebel:
@plribeiro3000 : Improved this a bit, mimics the real behavior more closely.

plribeiro3000:
It looks awesome. Thank you for taking the time to improve it!

plribeiro3000:
Thanks!

PR #15, opened 2015-12-03
Fix update_vm_interface

pedantic-git:
The previous implementation of update_vm_interface attempted to do a `ModifyVM_Task` using a new interface, freshly created with `create_interface()`.

On our environment, this caused the VM to be left with an unusable interface, with properties that I couldn't even examine without getting a 403 back from the server.

This version of update_vm_interface takes the *existing* raw interface object from the API and edits the network backing inside that object before submitting it back as a `ModifyVM_Task`.

This fixes the issue for us. Could you please take a look and decide if it belongs upstream?

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -33,7 +33,8 @@ module Fog
         end
 
         def get_vm_interface(vm_id, options={})
-          if raw = get_raw_interface(vm_id, options)
+          raw = get_raw_interface(vm_id, options)
+          if raw
             raw_to_hash(raw)
           else
             nil
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -29,31 +29,49 @@ module Fog
   #unitNumber: 7,
   #
         def list_vm_interfaces(vm_id, datacenter = nil)
-          get_vm_ref(vm_id, datacenter).config.hardware.device.grep(RbVmomi::VIM::VirtualEthernetCard).map do |nic|
-            {
-              :name    => nic.deviceInfo.label,
-              :mac     => nic.macAddress,
-              :network => nic.backing.respond_to?("network") ? nic.backing.network.name : nic.backing.port.portgroupKey,
-              :status  => nic.connectable.status,
-              :summary => nic.deviceInfo.summary,
-              :type    => nic.class,
-              :key     => nic.key,
-            }
-          end
+          get_raw_interfaces(vm_id, datacenter).map {|nic| raw_to_hash nic}
         end
 
         def get_vm_interface(vm_id, options={})
+          if raw = get_raw_interface(vm_id, options)
+            raw_to_hash(raw)
+          else
+            nil
+          end
+        end
+        
+        def get_raw_interfaces(vm_id, datacenter = nil)
+          get_vm_ref(vm_id, datacenter).config.hardware.device.grep(RbVmomi::VIM::VirtualEthernetCard)
+        end
+        
+        def get_raw_interface(vm_id, options={})
           raise ArgumentError, "instance id is a required parameter" unless vm_id
           if options.is_a? Fog::Compute::Vsphere::Interface
             options
           else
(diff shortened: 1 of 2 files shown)


lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
```diff
@@ -29,31 +29,49 @@ class Real
...
+        def get_raw_interface(vm_id, options={})
           raise ArgumentError, "instance id is a required parameter" unless vm_id
           if options.is_a? Fog::Compute::Vsphere::Interface
             options
           else
             raise ArgumentError, "Either key or name is a required parameter. options: #{options}" unless options.key? :key or options.key? :mac or options.key? :name
-            list_vm_interfaces(vm_id).find do | nic |
-              (options.key? :key and nic[:key]==options[:key].to_i) or (options.key? :mac and nic[:mac]==options[:mac]) or (options.key? :name and nic[:name]==options[:name])
+            get_raw_interfaces(vm_id).find do |nic|
+              (options.key? :key and nic.key==options[:key].to_i) or (options.key? :mac and nic.macAddress==options[:mac]) or (options.key? :name and nic.deviceInfo.label==options[:name])
```
plribeiro3000: Could you break this line for readability reasons please?

lib/fog/vsphere/requests/compute/modify_vm_interface.rb
```diff
@@ -19,11 +19,22 @@ def destroy_vm_interface(vmid, options = {})
         def update_vm_interface(vmid, options = {})
           raise ArgumentError, "instance id is a required parameter" unless vmid
 
-          interface = get_interface_from_options(vmid, options.merge(:server_id => vmid))
-          vm_reconfig_hardware('instance_uuid' => vmid, 'hardware_spec' => {'deviceChange'=>[create_interface(interface, interface.key, :edit)]})
+          interface = get_interface_from_options(vmid, options)
+          raw_interface = get_raw_interface(vmid, key: interface.key)
+          if options[:network]
+              interface.network = options[:network]
```
plribeiro3000: Could you use 2 spaces instead of tab?

lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
pedantic-git: Done.

lib/fog/vsphere/requests/compute/modify_vm_interface.rb
pedantic-git: Oops - sorry! My editor is set to 4-spaces by default and I keep forgetting to set it to 2 for these files.

plribeiro3000:
LGTM. @jrgarcia could you take a look to se if this code is ok?
I'm afraid i can't help more then review code. :sweat_smile:

jrgarcia:
@plribeiro3000 I'll take a look, but to be honest, I'm not too familiar with vSphere myself.

lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
```diff
@@ -29,31 +29,51 @@ class Real
...
-              :summary => nic.deviceInfo.summary,
-              :type    => nic.class,
-              :key     => nic.key,
-            }
-          end
+          get_raw_interfaces(vm_id, datacenter).map {|nic| raw_to_hash nic}
         end
 
         def get_vm_interface(vm_id, options={})
+          if raw = get_raw_interface(vm_id, options)
```
jrgarcia: Is this supposed to be an assignment? I usually prefer not to make assignments inside of a conditional's predicate.

plribeiro3000:
@jrgarcia Thats not a problem. Anything that you have makes you ahead of myself. If you have a login to try i out then is even better. =)

lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
pedantic-git: It is meant to be an assignment - what's your preferred way of saying "if x exists... do something with x"? This is the way I normally do it.

jrgarcia: I would normally do the assignment and test on separate lines. This is completely ok, but I see people run into issues all the time where they see `=` and read it as `==` inside an `if`.

pedantic-git: OK - I've changed this.

jrgarcia:
:+1: pending CI

PR #16, opened 2015-12-03
Add more mocks and add folder.destroy

pedantic-git:
Sorry this pull request is in two parts - unfortunately one depends on the other (although this can be rectified if required).

1. Adds/expands some more of the mocks (`create_folder`, `get_folder`, `list_datacenters`, `list_datastores`, `list_networks`, and several additions to `list_virtual_machines`). This doesn't break any of the tests except one that was using `servers.get` with a VM ID instead of a name - something that isn't supported by the real `servers.get`. We use these new mocks in our own tests but if you'd like them not to be merged upstream I can pull out the other part of this pull request separately.

2. Adds a `#destroy` to folder models and the associated `folder_destroy` request along with its mocks and a shindo test for it.

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -103,6 +103,7 @@ module Fog
       request :revert_to_snapshot
       request :list_processes
       request :upload_iso
+      request :folder_destroy
 
       module Shared
         attr_reader :vsphere_is_vcenter
@@ -329,7 +330,7 @@ module Fog
                  "tools_state"      => "toolsOk",
                  "name"             => "jefftest",
                  "operatingsystem"  => "Red Hat Enterprise Linux 6 (64-bit)",
-                 "path"             => "/",
+                 "path"             => "/Solutions/wibble",
                  "tools_version"    => "guestToolsUnmanaged",
                  "ipaddress"        => "192.168.100.187",
                  "uuid"             => "42329da7-e8ab-29ec-1892-d6a4a964912a",
@@ -340,6 +341,20 @@ module Fog
               },
               :datacenters => {
                 "Solutions" => {:name => "Solutions", :status => "grey", :path => ['Solutions']}
+              },
+              :folders => {
+                'wibble' => {
+                  'name' => 'wibble',
+                  'datacenter' => 'Solutions',
+                  'path' => '/Solutions/wibble',
+                  'type' => 'vm'
+                },
+                'empty' => {
+                  'name' => 'empty',
+                  'datacenter' => 'Solutions',
+                  'path' => '/Solutions/empty',
+                  'type' => 'vm'
+                }
               },
               :storage_pods =>
                 [{:id => "group-p123456",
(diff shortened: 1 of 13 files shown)


plribeiro3000:
@pedantic-git This PR blocks which one?

lib/fog/vsphere/compute.rb
```diff
@@ -329,7 +330,7 @@ def self.data
                  "tools_state"      => "toolsOk",
                  "name"             => "jefftest",
                  "operatingsystem"  => "Red Hat Enterprise Linux 6 (64-bit)",
-                 "path"             => "/",
+                 "path"             => "/Solutions/wibble",
```
plribeiro3000: I'm not into `vsphere` but does it make sense to do this change?

pedantic-git:
@plribeiro3000 It doesn't block any other PR but it has two distinct purposes (because the folder.destroy needs the new mocks for its test to pass).

lib/fog/vsphere/compute.rb
pedantic-git: This is in a mock. The `path` was not used by any other tests and I wanted to have a VM that was in a folder so the folder test works correctly. If you prefer, I could add another VM to the mocks instead of editing an existing one.

plribeiro3000: Oh. Got it. Thank you for the clarification. It is ok then. =)

lib/fog/vsphere/requests/compute/folder_destroy.rb
```diff
@@ -0,0 +1,31 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+        def folder_destroy(path, datacenter_name)
+          folder = get_raw_vmfolder(path, datacenter_name)
+          if folder
+            if folder.childEntity.size > 0
+              raise Fog::Vsphere::Errors::ServiceError, "Folder #{path} is not empty"
```
plribeiro3000: Could you convert this 2 nested ifs into 2 guard clauses?

lib/fog/vsphere/requests/compute/modify_vm_interface.rb
```diff
@@ -53,6 +53,13 @@ def destroy_vm_interface(vmid, options = {})
           raise ArgumentError, "interface is a required parameter" unless options and options[:interface]
           true
         end
+        
+        def update_vm_interface(vmid, options = {})
+          if options[:interface]
```
plribeiro3000: Could you convert this to a guard clause?

pedantic-git: Sure! Done.

lib/fog/vsphere/requests/compute/folder_destroy.rb
pedantic-git: Sure! Done.

plribeiro3000:
LGTM. @jrgarcia check this one as well please.

jrgarcia:
I don't have the ability to test this in a real environment at the moment, but it looks okay to me. :+1:

plribeiro3000:
Feel free to merge then. =)

PR #17, opened 2015-12-03
Additional create_vm options

pedantic-git:
This pull request adds the following options to the `create_vm` request:

1. `:annotation`: Allows setting of the notes field on a VM during creation.
2. `:boot_retry`: Set to a number of *ms*, will set the boot option to retry failed boots after that number of *ms*.
3. `:extra_config`: Allows passing of a hash to use as `extra_config` instead of the default one with `bios.bootOrder`.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -103,10 +103,18 @@ module Fog
         def boot_options attributes
           # NOTE: you must be using vsphere_rev 5.0 or greater to set boot_order
           # e.g. Fog::Compute.new(provider: "vsphere", vsphere_rev: "5.5", etc)
-          return unless @vsphere_rev.to_f >= 5
-          RbVmomi::VIM::VirtualMachineBootOptions.new(
-            :bootOrder => boot_order(attributes)
-          )
+          options = {}
+          if @vsphere_rev.to_f >= 5
+            options[:bootOrder] = boot_order(attributes)
+          end
+          
+          # Set attributes[:boot_retry] to a delay in miliseconds to enable boot retries
+          if attributes[:boot_retry]
+            options[:bootRetryEnabled] = true
+            options[:bootRetryDelay]   = attributes[:boot_retry]
+          end
+                   
+          options.empty? ? nil : RbVmomi::VIM::VirtualMachineBootOptions.new(options)
         end
 
         def boot_order attributes
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -7,6 +7,7 @@ module Fog
 
           vm_cfg        = {
             :name         => attributes[:name],
+            :annotation   => attributes[:annotation],
             :guestId      => attributes[:guest_id],
             :version      => attributes[:hardware_version],
             :files        => { :vmPathName => vm_path_name(attributes) },
@@ -260,12 +261,8 @@ module Fog
         end
 
         def extra_config attributes
-          [
-            {
-              :key   => 'bios.bootOrder',
-              :value => 'ethernet0'
-            }
-          ]
+          extra_config = attributes[:extra_config] || {'bios.bootOrder' => 'ethernet0'}
+          extra_config.map {|k,v| {:key => k, :value => v.to_s} }
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -106,10 +107,18 @@ def device_change attributes
...
+            options[:bootOrder] = boot_order(attributes)
+          end
+
+          # Set attributes[:boot_retry] to a delay in miliseconds to enable boot retries
+          if attributes[:boot_retry]
+            options[:bootRetryEnabled] = true
+            options[:bootRetryDelay]   = attributes[:boot_retry]
+          end
+                   
+          options.empty? ? nil : RbVmomi::VIM::VirtualMachineBootOptions.new(options)
```
plribeiro3000: I think you could do a guard clause at the beginning checking if options is empty and decrease the amount of code executed. This way it would not check for stuff we already know the result.

pedantic-git: `options` is a bit of a poor name, but this is the constructed boot options, not the `attributes` that's passed into the method. So it is always empty on line 110 and then depending on the two following if clauses it might gain values.

plribeiro3000: Oh. You are right.

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -7,6 +7,7 @@ def create_vm attributes = { }
 
           vm_cfg        = {
             :name         => attributes[:name],
+            :annotation   => attributes[:annotation],
```
plribeiro3000: Don't you need to add this attribute to the model?

pedantic-git: Good idea! I've done this and tested it works OK.

(I was only thinking about creating VMs, not querying them.)

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -106,10 +107,18 @@ def device_change attributes
...
-            :bootOrder => boot_order(attributes)
-          )
+          options = {}
+          if @vsphere_rev.to_f >= 5 and attributes[:boot_order]
+            options[:bootOrder] = boot_order(attributes)
+          end
+
+          # Set attributes[:boot_retry] to a delay in miliseconds to enable boot retries
+          if attributes[:boot_retry]
+            options[:bootRetryEnabled] = true
```
timogoebel: Could you also add this to the server model? :-)

plribeiro3000: Oh. thats true. Forgot to ask you to help out @timogoebel . =)

pedantic-git: I can add it to the model, but I was following the pattern of the :boot_order attribute which does not exist on the model.

Neither of them are trivial to add to the model because they don't correspond one-to-one with a property so they can't just go into `Fog::Compute::Vsphere::Shared#ATTR_TO_PROP`.

If it's important for them to be on the model, I can work around this limitation.

plribeiro3000: Whats exactly the limitation?

Maybe there is a better way to solve the problem that constant is solving.

pedantic-git: The limitation is that right now the attributes of the model are pulled in by the `convert_vm_view_to_attr_hash` function in compute.rb. This currently maps properties one-to-one with model attributes.

The two attributes `:boot_order` (this was already present) and `:boot_retry` (my parameter) construct more complex properties by expanding the single value given in Ruby to a more complex multi-value SOAP type. When retrieving a VM, it is this type that will come back and so if we wanted to store these values in the model, we would have to reverse the process of creating these types at VM creation time.

Since these are boot-related options, people are unlikely to need them beyond the creation of the VM. I'm assuming that's the reason the `:boot_order` was never added to the model and why I think perhaps it's not worth adding `:boot_retry` to the model either.

plribeiro3000: Got it.

But if the api returns those attributes i think it is nice to have them mapped in the model.

Not so long ago i've introduced the option `as` to the `attribute` DSL which is supposed to do the opposite of the option `aliases`.

It will map a simple attribute name like `boot` to something the api expects like `Boot`. But it only works for attribute mapping so far. If you need to transform one attribute into a structure then it does not work (yet).

Do you think it would fit the need here?

timogoebel: In my opinion, those options are definitely no requirement. However, they would be a nice to have. Maybe we should just open a task to add them at a later point.

pedantic-git: My gut feeling is that it's a fair amount of work for something that no one (so far) will be using. +1 for opening a task.

plribeiro3000: :+1: for a task. =)

pedantic-git: Since you both agree it should be a task, does that mean #17 is ready for merge now? Please let me know if there's anything more I can do.

PR #18, opened 2015-12-04
Add support for DRS Rules

pedantic-git:
This pull request adds support for DRS (affinity/anti-affinity) rules on a cluster. It adds:

* `cluster.rules`
* `cluster.rules.get`
* `cluster.rules.create` (or `new` &rarr; `vms=` &rarr; `save`)
* `rule.vms`
* `rule.destroy`

It's probably slightly controversial because it moves the `@connection` object into a shadow class to work around a longstanding bug in RbVmomi with respect to passing integers to the `xsd:any` type. (VMware expects an `xsd:int` but for some reason RbVmomi sends an `xsd:long`).

If you absolutely don't want the shadow class, I can take this code out, but I'll have to strip out the whole `destroy` functionality as this currently depends on passing the integer key to an `xsd:any`.

Let me know what you think!

### DIFF
--- a/lib/fog/vsphere/models/compute/rule.rb
+++ b/lib/fog/vsphere/models/compute/rule.rb
@@ -1,0 +1,46 @@
+module Fog
+  module Compute
+    class Vsphere
+      # ClusterRuleInfo
+      class Rule < Fog::Model
+        identity :key
+        
+        attribute :datacenter
+        attribute :cluster
+        attribute :name
+        attribute :enabled
+        # Type should be a class - either
+        #  - RbVmomi::VIM::ClusterAntiAffinityRuleSpec
+        #  - RbVmomi::VIM::ClusterAffinityRuleSpec
+        attribute :type
+        attribute :vm_ids
+        
+        def vms
+          vm_ids.map {|id| service.servers.get(id, datacenter) }
+        end
+        
+        def vms=(vms)
+          self.vm_ids = vms.map(&:instance_uuid)
+        end
+        
+        def save
+          requires :datacenter, :cluster, :name, :enabled, :type, :vm_ids
+          if vm_ids.length < 2
+            raise ArgumentError, "A rule must have at least 2 VMs"
+          end
+          if persisted?
+            raise "Update is not supported yet"
+          else
+            self.key = service.create_rule(attributes)
+          end
+          reload
+        end
+        
+        def destroy
+          service.destroy_rule(attributes)
+        end
+        
+      end
+    end
+  end
+end
--- a/lib/fog/vsphere/models/compute/rules.rb
+++ b/lib/fog/vsphere/models/compute/rules.rb
@@ -1,0 +1,30 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Rules < Fog::Collection
+        autoload :Rule, File.expand_path('../rule', __FILE__)
+        
+        model Fog::Compute::Vsphere::Rule
+        attribute :datacenter
+        attribute :cluster
+        
+        def all(filters = {})
+          requires :datacenter, :cluster
+          load service.list_rules(:datacenter => datacenter, :cluster => cluster)
+        end
+        
+        def get(key_or_name)
+          all.find {|rule| [rule.key, rule.name].include? key_or_name } or
+            raise Fog::Compute::Vsphere::NotFound, "no such rule #{key_or_name}"
+        end
+        
+        # Pass datacenter/cluster to every new rule
+        def new(attributes={})
+          requires :datacenter, :cluster
+          super(attributes.merge(datacenter: datacenter, cluster: cluster))
+        end
+
+      end
+    end
+  end
+end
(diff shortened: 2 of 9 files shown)


plribeiro3000:
@pedantic-git Since `RbVmomi` seems to be abandoned since 2012. What do you think of patching that specific class instead of doing a Shadow class here?

I can see that at some point we will have to take ownership of `RbVmomi` or change all requests to use somthing else anyway. =s

lib/fog/vsphere/models/compute/rules.rb
```diff
@@ -0,0 +1,28 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Rules < Fog::Collection
+        autoload :Rule, File.expand_path('../rule', __FILE__)
+        
+        model Fog::Compute::Vsphere::Rule
+        attr_accessor :datacenter
+        attr_accessor :cluster
```
plribeiro3000: Why didn't you use the `attribute` DSL here?

pedantic-git: Does that work for collection types? All the other collection types in fog-vsphere use `attr_accessor` and this is copied (primarily) from the `servers` type.

pedantic-git:
@plribeiro3000 Do you mean writing an entire monkeypatch for the `obj2xml` method inside fog-vsphere? My main concern about that is that [this is a massive chunk of code](https://github.com/rlane/rbvmomi/blob/master/lib/rbvmomi/connection.rb#L116) so I used a shadow class primarily to keep the intrusion into fog-vsphere to a minimum.

plribeiro3000:
:scream: OMG!

Didn't know it was that big.

You are right then. Your solution is far better for now.

lib/fog/vsphere/models/compute/rules.rb
plribeiro3000: Yeap. It does. The benefits is that you can use the `requires` method to tell a method needs that variable to no be nil.

pedantic-git: OK - I've switched the `rules` class to using attributes. You might want to open a task for someone to change all the other collections in fog-vsphere to match.

plribeiro3000:
@pedantic-git Could you take a look at the identation?
There are quite a few places where you've used 4 spaces.

pedantic-git:
@plribeiro3000 Curses! I think I've fixed the remaining ones in 29e70bc - please let me know if I missed any!

lib/fog/vsphere/models/compute/rule.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+        
+        def vms=(vms)
+          self.vm_ids = vms.map(&:instance_uuid)
+        end
+        
+        def save
+          requires :datacenter, :cluster, :name, :enabled, :type, :vm_ids
+          if vm_ids.length < 2
+              raise ArgumentError, "A rule must have at least 2 VMs"
```
plribeiro3000: There is one missing here. =)

lib/fog/vsphere/requests/compute/create_rule.rb
```diff
@@ -0,0 +1,46 @@
...
+  module Compute
+    class Vsphere
+      class Real
+        def create_rule(attributes={})
+          cluster = get_raw_cluster(attributes[:cluster], attributes[:datacenter])
+          # Check if it already exists and blow up if it does
+          # (otherwise ESX just happily accepts it and then considers it a conflict)
+          rule = cluster.configurationEx.rule.find {|n| n[:name] == attributes[:name]}
+          if rule
+              raise ArgumentError, "Rule #{attributes[:name]} already exists!"
```
plribeiro3000: Here again. :sweat_smile:

plribeiro3000:
@pedantic-git LGTM despite more 2 bad identend places

Dunno what you think but this `rbvomi` gem is kind of a mess imo. Is Vsphere Api too complex to handle on our own side?

pedantic-git:
@plribeiro3000 I've fixed those last two - sorry for my editor's settings!

The vSphere API is hideous - It's a massive hulking SOAP thing. Re-implementing the whole of fog-vsphere without RbVmomi would be a lot of work, I think, but it might be worth doing if a maintainer for RbVmomi can't be found!

jrgarcia:
We're talking about that here right now. Can't promise anything, but I'll update you guys with what we figure out about `rbvmomi`.

plribeiro3000:
:+1: for the changes.

Would it be too much if i asked you to squash the commits? :sweat_smile:

plribeiro3000:
@jrgarcia Lets us know what happens.

Perhaps we don't need to start over, just to continue `rbvomi`.

pedantic-git:
@plribeiro3000 It's not too much to ask, but I'm afraid I don't know how to do that with a pushed branch. Can you link me to a guide?

plribeiro3000:
Sure. If you know how to do it locally is pretty simple. Just squash it then do a `git push --force` and github will understand it and redo the commit tree right away. =)

Let me see if i can find you a link.

pedantic-git:
Ooh - I learnt something new today! I've done that. Let me know if there's anything else I can do!

plribeiro3000:
@pedantic-git Thats awesome! I always love when that happens. :heart: OSS :heart:

plribeiro3000:
LGTM!

PR #19, opened 2015-12-08
Pass folder information to more VM operations

pedantic-git:
Prior to this pull request, if you call `#vms` on a folder, the folder information is lost. This means:

* If you do `folder.vms.get(name)` it will actually fetch *every* VM in the datacenter and return a VM by that name wherever it lives, whether it's in the folder or not. This is both (incredibly) slow and leaky.
* If you do `folder.vms.create(hash)` you need to include datacenter and folder information in the hash even though they're already known by `folder`.

This pull request addresses these issues by changing this behaviour:

* `folder.vms.get(name)` will only return the VM if it is in the specified folder itself.
* `folder.vms(recursive: true).get(name)` will return the VM if it is in the specified folder or any of its children.
* `vms.create(hash)` will include datacenter/folder/cluster/resource_pool from the vms object if they are known (datacenter and folder are known in the case of `folder.vms` as a side-effect of the above).

The recursive search uses an `Enumerator` and `#find` to ensure we stop digging recursively once we've found the VM we're looking for, which provides a nice extra performance boost if your VM has a name that's early in the alphabet. :smile:

### DIFF
--- a/lib/fog/vsphere/models/compute/folder.rb
+++ b/lib/fog/vsphere/models/compute/folder.rb
@@ -10,9 +10,10 @@ module Fog
         attribute :path
         attribute :type
 
-        def vms
+        # Pass :recursive => true to get a Servers object that searches for VM names recursively
+        def vms(options = {})
           return [] if type.to_s != 'vm'
-          service.servers(:folder => path, :datacenter => datacenter)
+          service.servers(:folder => path, :datacenter => datacenter, :recursive => options[:recursive])
         end
 
         def to_s
--- a/lib/fog/vsphere/models/compute/servers.rb
+++ b/lib/fog/vsphere/models/compute/servers.rb
@@ -32,6 +32,17 @@ module Fog
         rescue Fog::Compute::Vsphere::NotFound
           nil
         end
+        
+        # Pass attributes we know about down to any VM we're creating
+        def new(attributes = {})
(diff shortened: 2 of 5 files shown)


plribeiro3000:
Hey @pedantic-git . sorry for the delay reviewing this one!

plribeiro3000:
Quick question. How bad would it be to always search recursively?

pedantic-git:
I don't think it would be too bad. Certainly I only use the recursive form in my code.

The main reason I implemented it this way was because the original interface to list_all_vms_in_folder and friends were non-recursive and I figured that's the functionality that was being used elsewhere by whoever created this code originally. It felt safer and less likely to break other people's code to add recursion as an option that's off by default.

plribeiro3000:
@pedantic-git You are right about that.
I was asking because if it makes sense we could do a major bump, where we can break old functionality.

But i think we can do this later. What do you think?

pedantic-git:
I think maybe it's best to leave it as an option. We might want a major bump anyway, because it's quite likely people are depending on the old (broken) functionality where folder.vms.get() would return VMs whether they were in the folder or not. I'll leave that decision up to you...

plribeiro3000:
Sure. =)

plribeiro3000:
Thank you for your time on this!

PR #20, opened 2015-12-10
Fix Fog::Mock.reset properly

jrgarcia:
- Remove the require of 'fog/vsphere/compute'
- Add module Compute and autoload Vsphere (previous autoload was reversed)


jrgarcia:
@plribeiro3000 Let me know if this works

plribeiro3000:
I'm sure it does @jrgarcia !

PR #22, opened 2016-01-14
storage pods: fix wrong string assignment

timogoebel:
I found a small, but very annoying bug in the storage pod code. Here is the fix.


plribeiro3000:
Wow. Great catch!

timogoebel:
@jrgarcia: Do you think you could release a new version of fog-vsphere? That would be greatly appreciated as it is blocking some foreman features right now.

jrgarcia:
@timogoebel Done! https://rubygems.org/gems/fog-vsphere/versions/0.5.0

PR #28, opened 2016-02-11
refactor storage pod handling

timogoebel:
Just a little refactoring for vm cloning to also check whether storage_pod is nil and not just if the key exists


jrgarcia:
:+1:

iNecas:
Without this patch, the `vm_clone` is broken as the original call for `get_storage_pod` has different number of arguments.  (see http://projects.theforeman.org/issues/14051). Would you mind releasing 0.6.1 with this, as it pretty limits the usage.

jrgarcia:
@iNecas Releasing now.

iNecas:
Thanks @jrgarcia !

PR #29, opened 2016-03-04
Expose numCoresPerSocket in vm_clone

iNecas:
We already support this option in `create_vm`. I've tested
with ESXi 5.1 and works fine.


jrgarcia:
:+1: Thanks @iNecas!

iNecas:
Thanks @jrgarcia ! :)

PR #34, opened 2016-05-12
Optionally process runcmd when creating a customspec from cloudinit

karmab:


plribeiro3000:
@jrgarcia Could you review that please?

jrgarcia:
@plribeiro3000 Looking now

jrgarcia:
@karmab So this section is only added to the Mock class, meaning this behavior only occurs when `Fog.mock? == true`. Is this what you were intending? What's the use case for this?

karmab:
well supposedly this code can be used when creating a vm, so it s not a mock right?

jrgarcia:
@karmab The section of the code that your change was added to is the mocking portion. Notice [the mock section](https://github.com/karmab/fog-vsphere/blob/082872c891fa8590d2096c2cf69f7024ed841265/lib/fog/vsphere/requests/compute/cloudinit_to_customspec.rb#L42) versus [the real section](https://github.com/karmab/fog-vsphere/blob/082872c891fa8590d2096c2cf69f7024ed841265/lib/fog/vsphere/requests/compute/cloudinit_to_customspec.rb#L4). When `Fog.mock? == true` the Mock class is used. When `Fog.mock? == false` the Real class is used.

karmab:
my bad, of course the code was intented to be run on real class... corrected it

jrgarcia:
@karmab Thanks for you contribution! After I get this merged in, I'll put out a new release with your change in it.

PR #38, opened 2016-06-15
list hosts in a cluster, deploy vm on a specific cluster

timogoebel:


jrgarcia:
Thanks again @timogoebel!

PR #40, opened 2016-07-11
Added logic to workaround the way that vsphere handles SCSI HBA

### LINKED ISSUE #39: Attempting to add more than 7 disks results in exception
Due to the way vsphere handles the SCSI ID reservation for the HBA, there is a mismatch between unit_numbers that fog uses and the actual SCSI ID for volumes above 7

The error produced is as follows:
```/Users/abohne/.rvm/gems/ruby-2.3.0/gems/fog-vsphere-0.8.0/lib/fog/vsphere/models/compute/volume.rb:83:in `save': undefined method `id' for nil:NilClass (NoMethodError)
	from ./create_oracle_vmdks.rb:107:in `block (3 levels) in <main>'
	from ./create_oracle_vmdks.rb:81:in `each'
	from ./create_oracle_vmdks.rb:81:in `block (2 levels) in <main>'
	from ./create_oracle_vmdks.rb:77:in `each'
	from ./create_oracle_vmdks.rb:77:in `block in <main>'
	from ./create_oracle_vmdks.rb:48:in `each'
	from ./create_oracle_vmdks.rb:48:in `<main>'```

### PR DESCRIPTION
**Pull request by abohne**
This fixes #39.

### CODE REVIEW

File: lib/fog/vsphere/models/compute/volume.rb [hunk 1]
```diff
@@ -49,7 +49,8 @@ def save
           requires :server_id, :size, :datastore
 
           if unit_number.nil?
-            used_unit_numbers = server.volumes.map { |volume| volume.unit_number }
+            # Vsphere maps unit_numbers 7 and greater to a higher SCSI ID since the pvscsi driver reserves SCSI ID 7
+            used_unit_numbers = server.volumes.map { |volume| volume.unit_number < 7 ? volume.unit_number : volume.unit_number - 1 }
```
plribeiro3000: If it is to a higher number as in your comment why are you decreasing by one instead? Shouldn't you be increasing by one if number is equal 7 or greater?

abohne: ```used_unit_numbers``` refers to the unit_numbers that fog has used.  When fog goes to add a volume with unit_number 7 for example, the SCSI ID ends up being 8.

Hope that provides some clarification.

plribeiro3000: So you decrease to calculate it right and later on you increase it back, right?

abohne: Yes, due to the way volumes are found in the following lines.

abohne: I don't see any of the other methods documented, how would you like this done?

plribeiro3000: I can be just an inline comment as you did but instead of being in two parts it could be at the top level of the method explaining how it works and why this workaround is in place.

abohne: OK I'll make that change

abohne:
Please check the updated inline comment and let me know if it is an acceptable explanation or not.

plribeiro3000:
Great rework here!

Thanks for bearing with me @abohne !

plribeiro3000:
@abohne Could you pin `json` version to `2.0.0` on ruby `1.9.3` an `jruby-19` please?

The most recent versions require at least ruby `2.0.0`.

abohne:
Sure, in ```gemfiles/Gemfile.1.9.2+```, correct?

plribeiro3000:
Hmmm. Just took a look now and what we have is wrong. We have `1.9.2-` which we are not using and `1.9.2` being used on `1.9.3` builds.

so maybe we want to remove the `1.9.2-` version and rename `1.9.2+` to `1.9` only. What do you think?

jrgarcia:
That's probably right. I followed the pattern of another `fog` project when I started, but we dropped support for 1.8.7 and honestly we probably shouldn't support < 1.9.3. I also think we should drop support for < 2.2 soon.

plribeiro3000:
@jrgarcia Sadly thats not something that we can do for now. We need to follow the mainstream `fog` which is gonna support ruby 2.2 for a while given some old linux distros that are LTS still use old ruby versions.

So for now we are good i believe.

abohne:
And to point another thing out, the Travis failures are occurring because ```rubocop``` dropped support for ruby 1.9.3 in 0.42.0.  If ```fog-vsphere``` wants to continue providing support for ruby 1.9.3, we would need to pin ```rubocop``` < 0.42.0 for that and jruby-19.

Please let me know how you would like me to proceed.

jrgarcia:
If you could do that as well, that would be great! If not, I can go ahead and merge your changes and fix the dependency issues. I apologize, I haven't had time to look into the Travis issues.

abohne:
I'll give it a shot.

plribeiro3000:
👍

jrgarcia:
👍 🎉  Thanks so much @abohne! I really appreciate your patience and willingness to fix the CI issues!

plribeiro3000:
Agree with @jrgarcia ! Thanks!

PR #43, opened 2016-08-18
Don't destroy stacktrace when things go wrong

ace13:
Currently, if anything gets caught by either of these handlers then the re-raised exception will be lacking the original stacktrace, which makes debugging the underlying cause impossible.

Example trace;
```
failed to create vm: NoPermission: Permission to perform this operation was denied.
   
RbVmomi::Fault: failed to create vm: NoPermission: Permission to perform this operation was denied.
  /usr/lib/ruby/gems/2.1.0/gems/fog-vsphere-1.0.0/lib/fog/vsphere/requests/compute/create_vm.rb:44:in `rescue in create_vm'
  /usr/lib/ruby/gems/2.1.0/gems/fog-vsphere-1.0.0/lib/fog/vsphere/requests/compute/create_vm.rb:8:in `create_vm'
  /usr/lib/ruby/gems/2.1.0/gems/fog-vsphere-1.0.0/lib/fog/vsphere/models/compute/server.rb:273:in `save'
  /usr/share/foreman/app/models/compute_resources/foreman/model/vmware.rb:364:in `create_vm'
  /usr/share/foreman/app/models/concerns/orchestration/compute.rb:82:in `setCompute'
  /usr/share/foreman/app/models/concerns/orchestration.rb:162:in `execute'
  /usr/share/foreman/app/models/concerns/orchestration.rb:107:in `block in process'
  /usr/share/foreman/app/models/concerns/orchestration.rb:99:in `each'
  /usr/share/foreman/app/models/concerns/orchestration.rb:99:in `process'
  /usr/share/foreman/app/models/concerns/orchestration.rb:35:in `on_save'
...
```

As can be seen, the trace ends on the rescue block where it's repackaged and re-raised.

This PR solves the issue by passing along the original exception backtrace into the repackaged exception object.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_folder.rb
+++ b/lib/fog/vsphere/requests/compute/create_folder.rb
@@ -13,7 +13,7 @@ module Fog
             # new path will be path/name, example: "Production/Pool1"
             new_folder.path.reject { |a| a.first.class == "Folder" }.map { |a| a.first.name }.join("/").sub(/^\/?Datacenters\/#{datacenter}\/vm\/?/, '')
           rescue => e
-            raise e, "failed to create folder: #{e}"
+            raise e, "failed to create folder: #{e}", e.backtrace
           end
         end
       end
--- a/lib/fog/vsphere/requests/compute/create_folder.rb
+++ b/lib/fog/vsphere/requests/compute/create_folder.rb
@@ -12,8 +12,6 @@ module Fog
             # output is cleaned up to return the new path
             # new path will be path/name, example: "Production/Pool1"
             new_folder.path.reject { |a| a.first.class == "Folder" }.map { |a| a.first.name }.join("/").sub(/^\/?Datacenters\/#{datacenter}\/vm\/?/, '')
-          rescue => e
-            raise e, "failed to create folder: #{e}", e.backtrace
           end
         end
       end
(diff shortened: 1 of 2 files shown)


plribeiro3000:
Hmmm. This makes me wonder. Do we need to rescue just to re-raise the exception?
Whats the value on that? Maybe we should jut let it explode and the final user would handle it?

ace13:
Considering the use case of Foreman, then the extended exception message gives a bit of context to the error that's finally reported to the web interface.  
Though it's not that much information that's added during the repackaging, so just letting the original exception fall through untouched would work just as well.

Should I write up another commit for that instead?

plribeiro3000:
Dunno. What would be the best scenario for foreman here?
Maybe what we need is to make foreman able to know whats wrong and do something about it. If thats the case i would say the solution here would be to not rescue anything in the part of the code and raise custom exceptions in the internal so any client (e.g foreman) would know better.
In the meantime we can leave it as is and have something like that in mind. What do you think?

ace13:
Usually when you want to do exception wrapping like this, then it's because you have access to additional information that could help diagnose the error. Like exception handling per line in a document you're parsing, in that case extending the error to `"#{exception} when parsing #{file}:#{line}"` would be helpful.

In this specific instance, I don't actually see anything that would be very relevant to provide as additional exception information, so if more informative custom exceptions are wanted then they might have to be done on a different level - a level where such information might be available.

plribeiro3000:
@ace13 I agree. =)

Removing this seems more reasonable to me. 👍

plribeiro3000:
Thanks for your contribution!

jrgarcia:
Thanks @ace13 and @plribeiro3000!

PR #44, opened 2016-08-23
Update fog-vsphere.gemspec

bentalbot:
Lock to the 1.8 version since 1.9 breaks on older Ruby

### DIFF
--- a/fog-vsphere.gemspec
+++ b/fog-vsphere.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 1.8.7'
 
   spec.add_runtime_dependency 'fog-core'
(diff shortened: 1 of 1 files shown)


jrgarcia:
@bentalbot I can fix this at some point today or tomorrow, but I actually plan to drop support for Ruby < 1.9.3. I'll probably accept this pull request and then create a new release of `fog-vsphere` and then release a newer version that drops support for Ruby < 1.9.3. Then, you could just pin your `.gemspec/Gemfile` to the last released version of `fog-vsphere` that support Ruby < 1.9.

jrgarcia:
In a pinch, for the time being, you can pin `rbvmomi` in your `Gemfile/gemspec`, but I'll release a new version with the pin and without 1.8.x support today. Sorry to break your build!

jrgarcia:
Thanks for the contribution @bentalbot! :tada:

PR #51, opened 2016-10-11
Do not fail on 'undefined method' when nicSettingMap not present

xprazak2:
When there is no `nicSettingMap` key in `custom_spec`, it results in:

```
| NoMethodError: undefined method `length' for nil:NilClass
| /home/vagrant/.rvm/gems/ruby-2.2.4/gems/fog-vsphere-1.2.0/lib/fog/vsphere/requests/compute/vm_clone.rb:520:in `vm_clone'
```

because `custom_nicSettingMap` is nil

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -509,15 +509,14 @@ module Fog
                 custom_adapter.primaryWINS = nic['adapter']['primaryWINS'] if nic['adapter'].key?('primaryWINS')
                 custom_adapter.secondaryWINS = nic['adapter']['secondaryWINS'] if nic['adapter'].key?('secondaryWINS')
                 custom_adapter.subnetMask = nic['adapter']['subnetMask'] if nic['adapter'].key?('subnetMask')
-                
+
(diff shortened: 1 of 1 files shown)


xprazak2:
You are certainly right about the append operation on an array. I did a little bit of digging and it seems to me that we do not need the offending code at all. Here's why:

If there is no nicSettingMap key in config, line [457](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/vm_clone.rb#L457) does not get executed and `custom_nicSettingMap` is not defined. Removing line 520 for this case takes care of the error.

If there is nicSettingMap key in config, `each` block on line [459](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/vm_clone.rb#L459) goes over adapter keys - adapter is required, there is at least one or [exception](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/vm_clone.rb#L455) is raised. Line [513](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/vm_clone.rb#L513) always creates adapter which is later appended to `custom_nicSettingMap`, so the array is never empty in this case and check is not necessary.

I also changed the comment, it was a bit misleading.

jrgarcia:
Thank you for your contribution!

PR #55, opened 2016-10-25
renaming the get_spec method in create_rule and create_group to provi…

carlospuga:
…de unique naming for the method and to avoid conflicting with one another.


plribeiro3000:
Nice catch. Thanks!

Could you please fix your git or github credentials so you can get ownership if those changes?

PR #56, opened 2016-10-25
changing modify_vm_controller so that it follows the changes made to …

carlospuga:
…create_controller in 1.4.0

### DIFF
--- a/lib/fog/vsphere/requests/compute/modify_vm_controller.rb
+++ b/lib/fog/vsphere/requests/compute/modify_vm_controller.rb
@@ -3,25 +3,13 @@ module Fog
     class Vsphere
       class Real
         def add_vm_controller(controller)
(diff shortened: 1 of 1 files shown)


plribeiro3000:
So this fixes a bug?

jrgarcia:
@plribeiro3000 I'm not sure exactly what happened, I may have missed something. It appears that the lines before the `vm_reconfig_hardware` are all handled in the `create_controller` method that is called on the `controller` that's passed in.

plribeiro3000:
@jrgarcia Thats fine. =)

I just wanted to make sure i was in the same page as @carlospuga.

jrgarcia:
Understandable. I'm still digging in and making sure this won't change any behavior.

plribeiro3000:
👍

PR #57, opened 2016-10-26
fixing modify_vm_volume so that it aligns with the changes in the sig…

carlospuga:
…nature of the create_disk method that were introduced in 1.4.0.  The create_disk method no longer requires the volume.unit_number as a parameter.


timogoebel:
Must have overlooked this, looks good to me.

jrgarcia:
Thanks again @carlospuga for your contributions! :tada:

PR #58, opened 2016-11-03
set default values more reliable

timogoebel:
fog-core sets these attributes "too late" and this can cause issues.


jrgarcia:
@timogoebel Can you try opening another PR. For some reason, it's saying I can't merge this because the base branch has been modified. I don't see anything wrong though.

timogoebel:
@jrgarcia : I have reset the branch to `origin/master` and cherry-picked my commit on top. Did that help?

PR #59, opened 2016-11-03
generate volume key reliably

timogoebel:
Generating the keys ourselves instead of relying on vsphere to do this solves errors when generating the boot order list.

### DIFF
--- a/lib/fog/vsphere/models/compute/volume.rb
+++ b/lib/fog/vsphere/models/compute/volume.rb
@@ -87,6 +87,17 @@ module Fog
           end
         end
 
+        def set_key
+          requires :controller_key, :unit_number
+
+          return unless key.nil?
+
+          # controller key is based on 1000 + controller bus
+          # disk key is based on 2000 + the SCSI ID + the controller bus * 16
+          controller_bus = controller_key - 1000
+          self.key = 2000 + (controller_bus * 16) + unit_number
+        end
+
         private
 
         def defaults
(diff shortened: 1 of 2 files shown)


timogoebel:
@jrgarcia : Thanks for taking a look. I don't think, this is an isolated Foreman problem. What breaks without this, is that you cannot create a vm with boot order set and scsi controllers of different kind. We just noticed it while working on an upstream feature.

timogoebel:
@jrgarcia : Don't want to nag or anything, just asking: Have you had a chance to take a look at this?

Let me explain, what this commit changes: Vsphere api expects all objects to have a unique identifier,  the `key`.  Before this commit, the keys for the volumes were generated randomly. But for setting the boot order you need to reference the volumes by key. This is obviously meant to break. So this commit moves the key generation to the volume model. When creating the volumes and the boot order, you can now use the same key.

jrgarcia:
Sorry! I got busy last week and was away for the weekend. I'll merge this and release a new version at some point today. Thanks!

timogoebel:
@jrgarcia : Thanks for taking care of both fog-vsphere and rbvmomi. You are doing a great job here.

jrgarcia:
Released [v1.5.1](https://rubygems.org/gems/fog-vsphere/versions/1.5.1) with changes

jrgarcia:
@timogoebel Thanks! I appreciate the kind words!

PR #66, opened 2017-01-17
impove folder and network mocks

timogoebel:


jrgarcia:
Thanks again for all your continued help @timogoebel!

PR #68, opened 2017-01-19
update volumes when server save method is called

TheKangaroo:
This PR extends the save method of server model to update changed volumes.
Related to #67


jrgarcia:
Thanks again, @TheKangaroo!

jrgarcia:
@timogoebel @TheKangaroo I've released v1.7.0 with the changes the two of you have made recently. Thanks so much for your PRs!

PR #70, opened 2017-01-31
Do not fail when cloning from template to different cluster

### LINKED ISSUE #69: Cloning VMs from a template to a different data store when resource pools are not in use
This is my first time opening an issue, so apologies in advance if this is lacking.

I am attempting to use The Foreman to provision VMs in VMware. When cloning a VM from a template on vSphere 6 when resource pools are not in use, unsure if this replicates itself on lower versions, the clone job will fail with 

> CannotAccessVmConfig: Unable to access the virtual machine configuration: Unable to access file [*target datastore for new VM here*]

I think the error comes from this code block in fog/vsphere/requests/compute/vm_clone.rb:
```ruby
if ( options.key?('resource_pool') && options['resource_pool'].is_a?(Array) && options['resource_pool'].length == 2 && options['resource_pool'][1] != 'Resources')
    cluster_name = options['resource_pool'][0]
    pool_name = options['resource_pool'][1]
    resource_pool = get_raw_resource_pool(pool_name, cluster_name, options['datacenter'])                                                                           
elsif ( vm_mob_ref.resourcePool == nil )
    # If the template is really a template then there is no associated resource pool,
    # so we need to find one using the template's parent host or cluster
    esx_host = vm_mob_ref.collect!('runtime.host')['runtime.host']
    # The parent of the ESX host itself is a ComputeResource which has a resourcePool
    resource_pool = esx_host.parent.resourcePool
    cluster_name = nil
end
```
When resource pools are not enabled (ie we are using the 'invisible' default pool 'Resources'), the first if statement is always false, because the pool is always named resources, and since we are cloning from a template, the elsif block is always true.
Because of this, whenever one attempts to clone a VM from a template to a cluster other than the one on which the template resides, the job will fail with the above error because the resource pool and cluster are set to those belonging to the template. 
I have inserted the following code block before the elsif statement:
```ruby
elsif ( options.key?('resource_pool') && options['resource_pool'].is_a?(Array) && options['resource_pool'].length == 2 && options['resource_pool'][1] == 'Resources') 
    cluster_name = options['resource_pool'][0]
    pool_name = options['resource_pool'][1]
    resource_pool = get_raw_resource_pool(nil, cluster_name, options['datacenter'])
```
This resolved the issue for me, as it sets the resource pool to that of the cluster we are cloning the VM to. I am unsure if this is completely correct, but I thought it might be helpful in some way.

01100010011001010110010101110000:
Resolves #69 by calling the target cluster resource pool, instead of the template's host's resource pool, when using the default resource pool.

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -115,6 +115,10 @@ module Fog
             cluster_name = options['resource_pool'][0]
             pool_name = options['resource_pool'][1]
             resource_pool = get_raw_resource_pool(pool_name, cluster_name, options['datacenter'])
+          elsif ( options.key?('resource_pool') && options['resource_pool'].is_a?(Array) && options['resource_pool'].length == 2 && options['resource_pool'][1] == 'Resources')
+            cluster_name = options['resource_pool'][0]
+            pool_name = options['resource_pool'][1]
+            resource_pool = get_raw_resource_pool(nil, cluster_name, options['datacenter'])
           elsif ( vm_mob_ref.resourcePool == nil )
             # If the template is really a template then there is no associated resource pool,
             # so we need to find one using the template's parent host or cluster


01100010011001010110010101110000:
No, I assigned that in case it was used elsewhere in the code and I had
missed it, but the code I added doesn't require it.
Also, I know the way I fixed this isn't exactly elegant, you're welcome to
change it however you see fit.

PR #74, opened 2017-02-20
send options to vm_migrate request

scambra:
options were ignored, except priority, so vm wasn't migrated


scambra:
I got vm_migrate request and migrate method in server model working, it fixes #75 (I wrote wrong issue number in commit)

scambra:
Yes, I have tested

jrgarcia:
Closes #75

PR #82, opened 2017-03-17
Add ability to detach volume from VM

scambra:
- Add detach to volume model
- Add remove_vm_volume as a request, it remove volume from vm without deleting file
- Refactored create_disk so it can be used for remove volume without deleting file, and reuse it in vm_reconfig_volumes request


scambra:
Rebased

PR #83, opened 2017-03-22
Add :connectable options to update_vm_interface

### LINKED ISSUE #81: Change connected status of interfaces
I would like to change connected status of interfaces in an existing vm. It's possible to change in vcenter, with edit settings for hardware. I don't see how to change it with fog, I read some sdk docs and I didn't find how to do it, but I guess it's possible.

If you give some guiding to do it with RbVmomi, I could give a try and send a pull request

scambra:
It allows to change :connected prop for interface, or
startConnected prop.

Using with server.update_interface should work:

    server.update_interface interface: interface, connectable: {connected: true, startConnected: true}

Fixes #81

### DIFF
--- a/lib/fog/vsphere/requests/compute/modify_vm_interface.rb
+++ b/lib/fog/vsphere/requests/compute/modify_vm_interface.rb
@@ -26,15 +26,16 @@ module Fog
             backing = create_nic_backing(interface, {})
             raw_interface.backing = backing
           end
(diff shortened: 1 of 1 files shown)


scambra:
I just tested with one vm and it works, thank you very much

jrgarcia:
@scambra Don't worry, I tested it before merging. I'm not _that_ crazy!

PR #84, opened 2017-04-05
validate connection before use

timogoebel:
This is a fix for `RbVmomi::Fault: NotAuthenticated:` when the connection gets stale. The code validates the vsphere connection before use and reconnects if the session expired.


jrgarcia:
Thanks so much for this @timogoebel! 🎉

PR #85, opened 2017-05-11
Properly escape DC names

ares:
This causes problems when DC name contains characters that have extra meaning in regular expressions e.g. parenthesis. The prefix is not matched and the full name contains also DC name. Later when we search for such cluster, it can't be found.


jrgarcia:
Thanks! I'll take a look at this and most likely merge this and release a new version with this change later today.

ares:
that would be very much appreciated

jrgarcia:
Just release [v1.9.2](https://rubygems.org/gems/fog-vsphere/versions/1.9.2) with your change. Thanks so much for the PR!

ares:
:fireworks: :+1:

PR #86, opened 2017-05-22
Rename vm

scambra:
Added a request to rename vm
It works changing name in server model and calling save, too.


jrgarcia:
By the way, I'm not sure if you saw it, but I released fog-vsphere v1.10.0 with this change in it yesterday.

scambra:
Thanks for warn me

PR #94, opened 2017-07-05
use portgroupkey to identify dv portgroup networks

### LINKED ISSUE #92: dvportgroup names should be referenced by `key`
See http://projects.theforeman.org/issues/19623 for back story and a possible patch.

Happy to submit one myself if you'd like.

(I am not the author of the above link, but, have come across the same problem)

wiad:
Resolves #92 by checking the network type and using the key attribute as ID if type is 'DistributedVirtualPortgroup'.

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_networks.rb
+++ b/lib/fog/vsphere/requests/compute/list_networks.rb
@@ -24,8 +24,14 @@ module Fog
         protected
 
         def network_attributes network, datacenter
+          raw_network = get_raw_network(network.name, datacenter)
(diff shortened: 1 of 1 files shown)


jrgarcia:
I'll plan to release a new version with this change later today. Thanks for the contribution! :tada:

jrgarcia:
@wiad I just pushed 1.11.1 with your contribution to [RubyGems](https://rubygems.org).

wiad:
@jrgarcia great, thats quick! Thank you for awesome work!

PR #98, opened 2017-07-26
vm clone: merge nics

timogoebel:
When cloning a VM from a template that has the boot order set, the operation may fail with the error `InvalidArgument: A specified parameter was not correct (configSpec.boot Options.boot Order)`. This is because the boot order setting references both storage volumes and network interfaces. During a clone operation (e.g. via Foreman), the network interface of the vm is replaced. The interface from the template is deleted and a new interface is added. The boot order reference is not updated and therefore invalid.

I did not find a way to actually modify the boot order of a vm during cloning, however I found a workaround for this: Instead of deleting the interface from the template and adding a new one, it's possible to update/edit the interface. Therefore the boot order reference doesn't get broken. This procedure is already done for volumes.

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -155,7 +155,7 @@ module Fog
             if options.key?('network_label')
               raise ArgumentError, "interfaces option can't be specified together with network_label"
             end
-            device_change.concat(modify_template_nics_specs(template_path, options['interfaces'], options['datacenter']))
+            device_change.concat(modify_template_nics_specs(vm_mob_ref, options['interfaces'], options['datacenter']))
           elsif options.key?('network_label')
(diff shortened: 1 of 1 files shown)


jrgarcia:
Sorry, I'm out this week and haven't had a chance to review this yet. Give me a couple of hours and I'll have it reviewed (and most likely) merged and released.

jrgarcia:
This looks good to me. I'll test it shortly just to verify (though I do trust you @chris1984 😄) and merge and release.

jrgarcia:
@timogoebel @chris1984 I just pushed up v1.11.3 with this change in it. Thank you for all of your contributions @timogoebel!

timogoebel:
Thanks @jrgarcia and @chris1984. We have opened a SR with VMWare to find out the reason why you cannot change/update the boot order during a clone operation. If we get this to work, we can probably revert this and add something like this:

```ruby
virtual_machine_config_spec.bootOptions = { :bootOrder => [...] }
```

PR #104, opened 2017-08-23
Fixed validation for power on/off requests (foreman)

scottd018:
Total n00b here to Git.  Done quite a bit of Ruby coding, so if I've done something incorrect as far as preferred Git process goes, please correct my behavior and I'll correct it the next time!

I found via Foreman that trying to power off an already powered off system, resulted in a 500 internal server error.  This is simple validation to not run the start/stop operations via RbVmomi if they are already in their desired state.


jrgarcia:
Thanks for your contribution!

Issue #79, opened 2017-02-27
Migrate volumes to different datastore


scambra:
I didn't see any request to change volume to different datastore, can you add it?

I checked SDK and I guess it's done with VirtualMachineRelocateSpec or VirtualMachineRelocateSpecDiskLocator, not sure which one

http://pubs.vmware.com/vsphere-65/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.vm.RelocateSpec.html
http://pubs.vmware.com/vsphere-65/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.RelocateSpec.DiskLocator.html

jrgarcia:
I'll take a look. I inherited this project, so I'm not entirely sure what's fully implemented and what's not. I'll take a look at that as soon as I get a chance though. I just started a new position today, so I may be a little bit slower to respond, but I'll do what I can as soon as possible.

jrgarcia:
Looking [here](https://github.com/fog/fog-vsphere/blob/930fc402cdcdf37e7614a44574ec8db833500baf/lib/fog/vsphere/requests/compute/vm_clone.rb#L62) and [here](https://github.com/fog/fog-vsphere/blob/930fc402cdcdf37e7614a44574ec8db833500baf/lib/fog/vsphere/requests/compute/vm_clone.rb#L586-L599), it looks like you may be able to use `vm_clone` to relocate a disk. I think if you pass a `transform` option to `vm_clone` that has the same properties as a `RelocateSpec`, you should be able to move it. Does that help at all?

scambra:
Thanks for giving a look. I'm not sure what you mean, I checked vsphere doc and it accepts flat and sparse strings. After reading this [doc](http://pubs.vmware.com/vsphere-65/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.vm.RelocateSpec.html) I think adding disk option to RelocateSpec in vm_clone would work to migrate only few volumes, vm_clone supports datastore option which I guess would change datastore for all volumes.

Although I was asking to migrate volume for an existing vm, without cloning, I was interested in migrate volumes when cloning. If not possible to migrate volumes for vm without cloning, I think I could clone vm and then drop old vm.

PR #105, opened 2017-08-28
add vm_relocate request which issues RelocateVM_Task

scambra:
I'm not sure how to add this to server model, or how to integrate in reconfig_volumes method of server model


scambra:
implements #79

PR #107, opened 2017-08-31
Support cloning volumes in different datastore

scambra:
add disk relocation spec to CloneVM_Task, so volumes from template can be relocated to different datastore

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -598,6 +598,7 @@ module Fog
               relocation_spec[:datastore] = datastore_obj
             end
           end
+          relocation_spec[:disk] = relocate_template_volumes_specs(vm_mob_ref, options['volumes'], options['datacenter'])
           # And the clone specification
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -768,6 +769,20 @@ def modify_template_volumes_specs(vm_mob_ref, volumes)
           specs.concat(new_volumes.map { |volume| create_disk(volume) })
           return specs
         end
+
+        def relocate_template_volumes_specs(vm_mob_ref, volumes, datacenter)
+          template_volumes = vm_mob_ref.config.hardware.device.grep(RbVmomi::VIM::VirtualDisk)
+          modified_volumes = volumes.take(template_volumes.size)
+          new_volumes      = volumes.drop(template_volumes.size)
```
jrgarcia: It doesn't appear that `new_volumes` is used anywhere. Is this correct? Is there a typo in the block below or this variable unnecessary?

scambra: You're right, I added new volumes to relocate spec, but it failed and also it isn't necessary, because they are created in right datastore, I will remove that line

PR #109, opened 2017-09-07
vm clone: generate new mac addresses

timogoebel:
https://github.com/fog/fog-vsphere/pull/98 introduced a regression that leads to duplicate mac addressed because the cloned host keeps the mac addresses of the parent host.
This forces the generation of new mac addresses during cloning.


jrgarcia:
I'll try and release this a little later today. Thanks for finding and fixing this, Timo!

PR #114, opened 2017-09-29
fix removing interfaces on vm_clone

scambra:
interface local variable does not exist, it is template_nic


jrgarcia:
Thanks @scambra!

jrgarcia:
I just released v1.13.1 with this fix.

PR #116, opened 2017-10-09
add requests to host

scambra:
shutdown host, enter and exit maintenance mode

### DIFF
--- a/lib/fog/vsphere/requests/compute/host_finish_maintenance.rb
+++ b/lib/fog/vsphere/requests/compute/host_finish_maintenance.rb
@@ -1,0 +1,23 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+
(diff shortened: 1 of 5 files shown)


chris1984:
Sorry for the late reply I will start testing these.

lib/fog/vsphere/requests/compute/host_finish_maintenance.rb
```diff
@@ -0,0 +1,16 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+
```
jrgarcia: Sorry to get back to you so late. Unfortunately, I've been swamped. This all looks good, but I would get rid of the  empty lines before and after the methods in the "requests" classes. Other than that it looks good to me.

scambra: removed empty lines

chris1984:
Looks good, Testing worked out well

PR #120, opened 2017-11-13
Fix regex issue when folder contains the datacenter name

fastlorenzo:
We have the following setup:

Datacenter: XYZ
Folder: ABC - XYZ
Cluster: Whatever

The regex tries to remove the datacenter name XYZ, but also remove part of the path which breaks the lookup of the resource pools later on.
By adding the ^ before the DC name in the regex, we ensure only that part is removed.


chris1984:
Thanks, @fastlorenzo for your contribution.

PR #122, opened 2018-01-12
ensure views are destroyed after use

timogoebel:
fog-vsphere uses Views to get data from vSphere. These views need to be closed manually or they will pile up and waste resources on the vCenter. This can cause problems when there are no more free resources for other services (eg. NSX Manager) that require them.


chris1984:
Looks good, tested here. Thanks @timogoebel

timogoebel:
Thanks for the review @chris1984!

PR #140, opened 2018-04-20
Require a network's datacenter for #update_vm_interface (#139)

### LINKED ISSUE #139: Modifying a virtual machine's interface requires the datacenter
The current implementation of `#update_vm_interface` at https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/modify_vm_interface.rb#L19 is not functioning.

The call to `create_nic_backing` requires the datacenter (silently) to be provided, otherwise the network will not be found an a nil error will be thrown from `#list_container_view`.

```ruby
def update_vm_interface(vmid, options = {})
  raise ArgumentError, 'instance id is a required parameter' unless vmid

  interface = get_interface_from_options(vmid, options)
  raw_interface = get_raw_interface(vmid, key: interface.key)
  if options[:network]
    interface.network = options[:network]
    backing = create_nic_backing(interface, {})
    raw_interface.backing = backing
  end
  apply_options_to_raw_interface(raw_interface, options)
  spec = {
    operation: :edit,
    device: raw_interface
  }
  vm_reconfig_hardware('instance_uuid' => vmid, 'hardware_spec' => { 'deviceChange' => [spec] })
end
```

leewaa:
Fixes #139

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -43,11 +43,15 @@ module Fog
 
         def get_raw_interface(vm_id, options = {})
           raise ArgumentError, 'instance id is a required parameter' unless vm_id
+
           if options.is_a? Fog::Compute::Vsphere::Interface
             options
+
           else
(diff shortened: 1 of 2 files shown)


leewaa:
So this should resolve the issue with modifying a vm's network interface. I also checked if these changes have any other implications and could not spot anything.

timogoebel:
Thanks, @leewaa.

PR #149, opened 2018-06-29
Fixes #148 - Add unclustered ESXI hosts back to cluster selection.

### LINKED ISSUE #148: Non-clustered ESX hosts, no longer appear in host creation
We are seeing a new issue with Foreman 1.17 and ESX 6.5, that previously worked on Foreman 1.16.

ESX hosts that are non clustered, no longer show up as an option in the Virtual Machine tab. It used to be that you could choose any host within an ESX DataCenter, regardless if it was part of a cluster or not. 
We have numerous vcenters, many with dozens of hosts that are non-clustered.

It would be nice if the previous methodology worked, as it supported both clustered and non-clustered ESX hosts.

chris1984:


chris1984:
@timogoebel if this looks good can we get it into the version you are about to make.

timogoebel:
Thanks, @chris1984.

PR #152, opened 2018-07-20
Add server guard to unit_number calculation

ares:


timogoebel:
Ha, I just typed almost the exact thing. :-)

ares:
let me know if you prefer to keep it in separate PR

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -344,7 +344,14 @@ def initialize_interfaces
 
         def initialize_volumes
           if attributes[:volumes] && attributes[:volumes].is_a?(Array)
-            attributes[:volumes].map! { |vol| vol.is_a?(Hash) ? service.volumes.new({ server: self }.merge(vol)) : vol }
+            attributes[:volumes].map! do |vol|
+              if vol.is_a?(Hash)
+                service.volumes.new({server: self}.merge(vol))
```
timogoebel: @ares: Do you mind adding two spaces here so rubocop is happy?
`{server: self}` -> `{ server: self }`

PR #153, opened 2018-08-13
Add NSX-T support in create_vm.rb

yasensim:
This allows connecting VMs to NSX-T Logical Switches. Please merge.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -171,6 +171,12 @@ module Fog
                 switchUuid: raw_network.config.distributedVirtualSwitch.uuid
               )
             )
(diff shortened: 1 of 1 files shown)


timogoebel:
LGTM.

nicholaspearson:
Thanks @yasensim 

Would appreciate if we could pull a support release too for fog-vsphere-0.6.0 with this code in. Unfortunately we are slightly version locked for the moment at 0.6.0.

Is this possible?

yasensim:
Makes sense @dixonly I just changed it.
@nicholaspearson I will let the maintainers to advice s create_vm.rb is very different in v1.6.0. I can send separate PR.

timogoebel:
If you could open a separate PR with the changes for 1.6.0 that‘d definitely help.
I‘m on pto this week and far away from a real computer, but will gladly create a release next week.

PR #157, opened 2018-10-02
Help prevent hard fail on nil undefined legacy networks.

chris1984:
PR to work to prevent this with a hard fail:

https://bugzilla.redhat.com/show_bug.cgi?id=1612921

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -61,11 +61,19 @@ module Fog
 
         private
 
+        # rubocop:disable Style/ConditionalAssignment
         def raw_to_hash(nic)
+          if nic.backing.respond_to?(:network)
+            network = nic.backing.network.name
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
```diff
@@ -62,10 +62,17 @@ def get_raw_interface(vm_id, options = {})
         private
 
         def raw_to_hash(nic)
+          if nic.backing.respond_to?('network')
```
timogoebel: While we're at it, can you please change `'network'` to a symbol (`:network`)? Same for port below.

chris1984:
@timogoebel thanks for the review :) Do you want me to change it to make rubopcop happy or want me to add a line to disable the cop for this if block and renable it after? The latter would be easier

chris1984:
@timogoebel I went with the latter, updated if you want to merge.

timogoebel:
Thanks, @chris1984.

PR #159, opened 2018-10-03
speed up datastore loading

timogoebel:
@chris1984: The same for datastores. This reduces load times from ~50s to ~4s.

From my findings, Folders and Datastores take the longest in Foreman.


chris1984:
Thanks, @timogoebel going to merge since I am ok with the method even though codeclimate does not like it. Travis passed and that makes me happy.

PR #163, opened 2018-11-06
fix adding 8th volume to vm (fixes #146)

### LINKED ISSUE #146: Add 8th volume fails
Exception is raised when filling id and other attributes after creating in vmware, but volume is actually created.

Here is a extract of save method:

```ruby
        def save
          raise Fog::Errors::Error.new('Resaving an existing object may create a duplicate') if persisted?
          requires :server_id, :size, :datastore

          set_unit_number

          data = service.add_vm_volume(self)

          if data['task_state'] == 'success'
            if self.unit_number >= 7
              self.unit_number += 1
            end

            # We have to query vSphere to get the volume attributes since the task handle doesn't include that info.
            created = server.volumes.all.find { |volume| volume.unit_number == self.unit_number }
```

Save method calls set_unit_number, which calls calculate_free_unit_number. That method avoids setting unit_number to 7, but after volume is created, unit_number is increased by 1 and created volume is not found.

Is it this code needed? Because I think it's breaking adding 8th volume:

```ruby
            if self.unit_number >= 7
              self.unit_number += 1
            end
```

scambra:
Don't increment unit_number if 7 or higher, volume won't be found, set_unit_number already avoids setting unit_number to 7.

### DIFF
--- a/lib/fog/vsphere/models/compute/volume.rb
+++ b/lib/fog/vsphere/models/compute/volume.rb
@@ -60,8 +60,6 @@ module Fog
           data = service.add_vm_volume(self)
 
           if data['task_state'] == 'success'
-            self.unit_number += 1 if unit_number >= 7
-
             # We have to query vSphere to get the volume attributes since the task handle doesn't include that info.
             created = server.volumes.all.find { |volume| volume.unit_number == self.unit_number }
 


chris1984:
Thanks @scambra

PR #164, opened 2018-11-06
Relocate

scambra:
add option to relocate vm's datastore
Also I added support for diskBackingInfo in VirtualMachineRelocateSpecDiskLocator, I needed that option when migrating volumes between datastores of different type.


scambra:
I had removed changes in pool option, I don't remember why I did them, probably just copied from vm_clone, I didn't test those changes because I'm not using resource pools, so I guess it's better to leave as it was

chris1984:
@timogoebel can you review the code and I will take care of doing regression testing with Foreman before we merge?

PR #165, opened 2018-11-08
add required datacenter param to all calls which may use get_vm_interface

scambra:
Calling server.interfaces.get('mac_address') failed because datacenter is required
It could fail calling server.update_vm_interface because datacenter is required,
or calling interface.destroy

### DIFF
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -230,6 +230,7 @@ module Fog
 
         def update_interface(attrs)
           wait_for { !ready? } if interface_ready? attrs
+          attrs = attrs.merge(datacenter: datacenter) unless attrs.key? :datacenter
           service.update_vm_interface(id, attrs)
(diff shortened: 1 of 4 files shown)


chris1984:
@scambra thanks I will start testing this.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -230,6 +230,7 @@ def add_interface(attrs)
 
         def update_interface(attrs)
           wait_for { !ready? } if interface_ready? attrs
+          attrs = attrs.merge(datacenter: datacenter) unless attrs.key? :datacenter
```
timogoebel: You can use `attrs[:datacenter] = datacenter` here. That avoids the call to `merge`.

scambra: changed

chris1984:
Thanks, @scambra for all the contributions.

PR #167, opened 2018-11-12
patch attribute[:datacenter] for create interface

kensoncheung:
This patch is for working around create VM interface failure due to datacenter undefined. (undefined method `networkFolder' for nil:NilClass)


chris1984:
@kensoncheung thanks, will start regression testing on Foreman and normal testing.

chris1984:
Tests failing because of rubocop, not functionality, will make an issue to update rubocop.

PR #171, opened 2018-11-14
Support getting folder types other than vm folder

kensoncheung:
Enhanced get_folder() and get_raw_folder() implementation to support folder types other than 'vm'


kensoncheung:
@chris1984 Thanks!

PR #180, opened 2018-11-27
Fixes #179 - Update vm_clone template to not always be false.

### LINKED ISSUE #179: vm_clone: suggest to modify the :template parameter as an options['template'] instead of false in CloneVM_Task
The :template parameter is always false in the current implementation. However, I have a use case which is to clone a VM to a template. Therefore, I suggest to modify this field as an option. Below is the code suggestion:
```
clone_spec = RbVmomi::VIM.VirtualMachineCloneSpec(:location => relocation_spec,
                                                            :config => virtual_machine_config_spec,
                                                            :customization => customization_spec,
                                                            :powerOn  => options.key?('power_on') ? options['power_on'] : true,
                                                            :template => options.key?('template') ? options['template'] : false)
```

chris1984:
@manchiu do you want to review/test this for me, while I take care of the Foreman tests?

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -641,7 +641,7 @@ module Fog
                                                             config: virtual_machine_config_spec,
                                                             customization: customization_spec,
                                                             powerOn: options.key?('power_on') ? options['power_on'] : true,
-                                                            template: false)
+                                                            template: options.key?('template') ? options['template'] : false)
 
           # Perform the actual Clone Task
           # Clone VM on a storage pod


lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -637,11 +637,11 @@ def vm_clone(options = {})
             end
           end
           # And the clone specification
-          clone_spec = RbVmomi::VIM.VirtualMachineCloneSpec(location: relocation_spec,
-                                                            config: virtual_machine_config_spec,
-                                                            customization: customization_spec,
-                                                            powerOn: options.key?('power_on') ? options['power_on'] : true,
-                                                            template: false)
+          clone_spec = RbVmomi::VIM.VirtualMachineCloneSpec(:location => relocation_spec,
```
timogoebel: Can we please use the newer ruby syntax here?

chris1984:
@timogoebel does that look better?

manchiu:
> @manchiu do you want to review/test this for me, while I take care of the Foreman tests?

Sure. @chris1984. Just reviewed and tested. I think it is good enough to proceed. Thanks for the patch!

chris1984:
Thanks @manchiu @timogoebel for the testing and code review,  all the regression tests pass, going to merge.
```
Image:

Normal VM creation with single HDD on Local Datastore - Passed
Normal VM creation with added HDD on Local Datastore - Passed
Normal VM creation with single HDD on Storage Pod - Passed

PXE:

Normal VM creation with single HDD on Local Datastore - Passed

Normal VM creation with added HDD on Local Datastore - Passed

Normal VM creation with single HDD on Storage Pod - Passed

Normal VM creation with added HDD on Storage Pod - Passed

Bootdisk:

Normal VM creation with added HDD on Local Datastore - Passed

Normal VM creation with single HDD on Storage Pod - Passed

Normal VM creation with added HDD on Storage Pod - Passed


Network:

Normal VM creation with NIC in portgroup on Standard Switch - Passed

Normal VM creation with NIC in portgroup on Distributed Switch - Passed

Normal VM creation with NIC in port group on Distributed Switch with VLAN - Passed

Normal VM creation with additional NIC - Passed
```

PR #183, opened 2018-12-09
Fixes #25584 - Raise error in case path and cluster doesn't exists

shiramax:


chris1984:
Thanks! I will start working on testing this

Issue #182, opened 2018-11-30
Unable to load Fog::Vsphere::Compute


serj1408:
How to avoid these type of warnings? It used to work and I can't find any new format documentation. Or is it a bug?
```
require 'fog/vsphere'
credentials = {
    :provider=>"vsphere",
    :vsphere_username=>ENV['VCENTER_USERNAME'],
    :vsphere_password=>ENV['VCENTER_PASSWORD'],
    :vsphere_server=>"xxxxx",
    :vsphere_expected_pubkey_hash => "xxxxx"
}
compute = Fog::Compute.new(credentials)

[fog][DEPRECATION] Unable to load Fog::Vsphere::Compute
[fog][DEPRECATION] The format Fog::Compute::Vsphere is deprecated
```
Ruby version 2.1.3
```
    fog-core (2.1.2)
      builder
      excon (~> 0.58)
      formatador (~> 0.2)
      mime-types
    fog-vsphere (2.5.0)
      fog-core
      rbvmomi (~> 1.9)
```

timogoebel:
This might be due to a change in fog-core.
I believe we need to change our code to use autoload, similar to https://github.com/fog/fog-digitalocean/commit/5a023b40ee971529e2d3d5b8fb27869f5fb01934.

@chris1984: We have seen this in Foreman as well, so we should take care of this. Better sooner than later ;-)

tarcinil:
@chris1984 - Have you completed any work on this?  Any issues with me making an attempt at it?

timogoebel:
cc: @ares, @ezr-ondrej

chris1984:
@tarcinil have at it

ezr-ondrej:
@tarcinil I would do what @timogoebel suggested and move to the autoload structure of the project. Have you done anything yet? Or should I do it?

tarcinil:
@ezr-ondrej I have not started the work but that was my plan.  I was planning on starting my work tonight and having something reviewable by the weekend.

ezr-ondrej:
If you'll give me about an hour, I'll produce testable version of that (WIP PR), I am finishing the testing. So you can base on that. I will move the files and give it some spins to see if it is not marginally broken.

ezr-ondrej:
After bit of an investigation I beleive the prefered way is the oposite of what **digitalocean** did.
And I believe the finite answer is: https://github.com/fog/fog-core/blob/master/spec/compute_spec.rb.

ares:
[...]

PR #190, opened 2019-03-18
Refs #182 - use autoload

ezr-ondrej:
Moves the compute files to correct path for autoloading.

I have moved the files, tested on foreman:
- Listing VMs
- Create VM (one disk, one nic)
- Delete VM


tests/models/compute/server_unit_tests.rb
```diff
@@ -1,6 +1,6 @@
 require_relative '../../helper'
 
-require 'fog/vsphere/models/compute/server'
+# require 'fog/compute/vsphere/models/server'
```
timogoebel: Can this line be removed?

ezr-ondrej: I believe so, can not confirm though as the tests are broken. Would be maybe good idea to make them pass first and then continue on this one?

timogoebel: That sounds like an excellent idea.

ezr-ondrej:
I would like to close it in favor of #192

PR #191, opened 2019-03-18
Tests passing

ezr-ondrej:
It would deserve to refactor them more, but I believe, we should switch to minitest as the core did anyway, so I believe now it is about at least passing.


timogoebel:
Thanks, @ezr-ondrej!

PR #192, opened 2019-03-18
Use advised naming

ezr-ondrej:
Namespacing prefered according to https://github.com/fog/fog-core/blob/master/spec/compute_spec.rb

After few necessary changes in foreman i tested:
- List VMs
- Delete VM
- Power On VM (power off doesnt work in my ESXi for some reason so untested).
- Create VM

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Or install it yourself as:
 gem install fog-vsphere
 ```
 
+## Upgrading
+
+In version 3.0.0 we have changed the namespacing of `Compute` service.
+`Fog::Compute::Vsphere` became `Fog::Vsphere::Compute` as recommended by fog-core.
+We hope we are not causing any issues as `Fog::Compute[:vsphere]` is only public API.
+
 ## Usage
 
 To connect to your vSphere instance with Fog vSphere:
@@ -67,7 +73,7 @@ compute.list_virtual_machines
 
 # List a VM's SCSI controllers
 compute.list_vm_scsi_controllers('ab589f9a-af35-428e-9690-9b96587d86f3')
-#=> [<Fog::Compute::Vsphere::SCSIController
+#=> [<Fog::Vsphere::Compute::SCSIController
 #=>   shared_bus='noSharing',
 #=>   type='VirtualLsiLogicController',
 #=>   unit_number=7,
--- a/lib/fog-vsphere.rb
+++ b/lib/fog-vsphere.rb
@@ -1,0 +1,1 @@
+require 'fog/vsphere'
--- a/lib/fog/vsphere.rb
+++ b/lib/fog/vsphere.rb
@@ -23,8 +23,4 @@ module Fog
       nil
     end
   end
-
-  module Compute
-    Vsphere = ::Fog::Vsphere::Compute
-  end
 end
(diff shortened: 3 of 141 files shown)


lib/fog/vsphere.rb
```diff
@@ -25,4 +23,8 @@ def self.class_from_string(name, default_path = '')
       nil
     end
   end
+
+  module Compute
+    Vsphere = ::Fog::Vsphere::Compute
```
ezr-ondrej: not sure about this nasty alias.

timogoebel: Neither am I. Do others have this?
With this PR merged I‘d like to bump the major version. So I guess we don‘t need to stay backwards compatible.

timogoebel: Let's fix this once and for all. We'll bump the major version that includes this commit. For Foreman this should not be a problem at all (we might just need to restrict the gem version via a commit in the stable branches).

@tbrisker: Opinions?

ezr-ondrej: I was just worried about the even older versions, but I guess it is 1) not our problem 2) not an issue, because we will require higher version of fog in new version, right?

tbrisker has a party today :balloon: so won't answer till tomorrow probably.

ezr-ondrej:
Reverted the alias. That case, we should put it in README probably, for fast debugging.

timogoebel:
Yeah, we should update the README with what changed in 3.0.0. :-)

ezr-ondrej:
Should I do that, or we will do that in follow-up (bumb PR)?

timogoebel:
@ezr-ondrej: It can't hurt to add that now. :-)

lib/fog/vsphere/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module Vsphere
-    VERSION = '2.5.0'.freeze
+    VERSION = '3.0.0'.freeze
```
ezr-ondrej: I have bumped.

lib/fog-vsphere.rb
```diff
@@ -0,0 +1 @@
+require 'fog/vsphere'
```
ezr-ondrej: I have add high level require

README.md
```diff
@@ -26,6 +26,12 @@ Or install it yourself as:
 gem install fog-vsphere
 ```
 
+## Upgrading
```
ezr-ondrej: I have add upgrading notice

ezr-ondrej:
@chris1984 It would be perfect, but I guess it doesn't make sence to run them on foreman since this is going to break foreman for sure after bump, could I ping you on a PR in foreman, once this is merged for the tests, please?

README.md
```diff
@@ -26,6 +26,12 @@ Or install it yourself as:
...
+
+In version 3.0.0 we had changed the namespacing of Compute.
```
timogoebel: nit: `have changed`
@chris1984: Can you check the English here please?

lib/fog/vsphere.rb
tbrisker: i'm guessing this is no longer relevant, but whatever you folks think is best for the gem is fine by me :)

timogoebel:
Thanks a bunch, @ezr-ondrej!

PR #195, opened 2019-03-23
Add boot_order create tests

ezr-ondrej:
I made a try for real compute class testing - I feel that it is quite important for the requests testing, but bit difficult as clearly visible from amout of mockstubing compared to actual testing.
Got me thinking if we couldn't abstract every command to it's class, but not enough motivation to try by my own without some aprior consensus, so I've tried to push forward at least like this :)

### DIFF
--- a/tests/helper.rb
+++ b/tests/helper.rb
@@ -5,5 +5,6 @@ require 'mocha/api'
 Shindo::Tests.send(:include, Mocha::API)
 
 require 'minitest/autorun'
+require 'mocha/minitest'
 
 require File.expand_path('../../lib/fog/vsphere', __FILE__)
--- a/tests/requests/compute/create_vm_tests.rb
+++ b/tests/requests/compute/create_vm_tests.rb
@@ -1,0 +1,89 @@
+require_relative '../../helper'
+
+require 'fog/vsphere/requests/compute/create_vm'
+
+describe Fog::Vsphere::Compute::Real do
+  let(:connection) { mock('RbVmomi::VIM') }
+
+  before do
+    Fog.unmock!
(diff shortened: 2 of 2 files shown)


timogoebel:
+1 for adding tests, especially minitest. We definitely need more here to be more confident we don't break stuff. The mocking is tough to understand. Have you considered using `webmock` to stub the rbvmomi calls?

> Got me thinking if we couldn't abstract every command to it's class,

I don't follow. Please explain. :-)

ezr-ondrej:
> +1 for adding tests, especially minitest. We definitely need more here to be more confident we don't break stuff. The mocking is tough to understand. Have you considered using `webmock` to stub the rbvmomi calls?

+1 for webmock, probably bit cleaner solution, but doesn't solve that load of dependencies we need to stub before getting on the real testing. I believe it says something about how deeply coupled is actually the code. It would be one webmock per one stub here anyway.

> I don't follow. Please explain. :-)
If we could define class per each request command, for not polluting the `Fog::Vsphere::Compute::Real`.
Something like: https://gist.github.com/ezr-ondrej/2b194609864cf10c7de5d650ad9f168b
I believe it would help to cleanup the code a bit and the tests would became much simpler as we would test decoupled code.

timogoebel:
> If we could define class per each request command, for not polluting the `Fog::Vsphere::Compute::Real`.
> Something like: https://gist.github.com/ezr-ondrej/2b194609864cf10c7de5d650ad9f168b
> I believe it would help to cleanup the code a bit and the tests would became much simpler as we would test decoupled code.

Yep, that sounds like the best solution to me.

Issue #197, opened 2019-03-24
Encapsulate requests


ezr-ondrej:
As discussed in https://github.com/fog/fog-vsphere/pull/195, it would be nice not to pollute class `Fog::Vsphere::Compute::Real` that much and encapsulate each request in separate request class.
I'll try to do that and tag this issue for that.

PR #198, opened 2019-03-24
[WIP] Refs #197 - Encapsulate create_vm command

ezr-ondrej:
Follow up of #195 - the request encapsulation.
This is a draft how this would look like.

I have taken the create_vm command, I have put it in separate class.
Pulled all dependencies into its ancestor.
The ancestor is quite big now, what would have to be addressed.
But I believe it is still better than having the same in `Fog::Vsphere::Compute::Real` class.
After some improvements agreed upon, I will rewrite #195 on top of it.

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -723,27 +723,11 @@ module Fog
         end
 
         def list_container_view(datacenter_obj_or_name, type, container_object = nil)
-          dc = if datacenter_obj_or_name.is_a?(String)
-                 find_raw_datacenter(datacenter_obj_or_name)
-               else
-                 datacenter_obj_or_name
-               end
+          shared_request.send('list_container_view', datacenter_obj_or_name, type, container_object)
+        end
 
-          container = if container_object
-                        dc.public_send(container_object)
-                      else
-                        dc
-                      end
-
-          container_view = connection.serviceContent.viewManager.CreateContainerView(
-            container: dc,
-            type: [type],
-            recursive: true
-          )
-
-          result = container_view.view
-          container_view.DestroyView
-          result
+        def shared_request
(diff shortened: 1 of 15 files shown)


lib/fog/vsphere/requests/request.rb
```diff
@@ -0,0 +1,19 @@
+module Fog
+  module Vsphere
+    module Requests
+      class Request
```
ezr-ondrej: This is base class for request, after some flame or fame, I would try to suggest it to `fog-core` as I believe in principle this is a better pattern.

lib/fog/vsphere/requests/compute/request.rb
```diff
@@ -0,0 +1,166 @@
+require 'fog/vsphere/requests/request'
+
+module Fog
+  module Vsphere
+    module Requests
+      module Compute
+        class Request < Fog::Vsphere::Requests::Request
```
ezr-ondrej: This file is the worst place of whole thing.
{Not mentioning the naming as I expect discussion about that.}
It creates a dependency cointainer, but I believe we could narrow it down by the commands calling one another, once we encapsulate them all, as that is what whole file basically does now anyways.
And I kind of like the idea as it would be more explicit dependency.
Unlike now, when we call other requests private methods.

lib/fog/vsphere/compute.rb
```diff
@@ -702,27 +702,11 @@ def validate_ssl_connection
...
-            recursive: true
-          )
+          shared_request.send('list_container_view', datacenter_obj_or_name, type, container_object)
+        end
 
-          result = container_view.view
-          container_view.DestroyView
-          result
+        def shared_request
+          @shared_request ||= Fog::Vsphere::Requests::Compute::Request.new(connection)
```
ezr-ondrej: This is what I came up with for backward compatibility.
It should be removed after the dependency container is unwrapped.

PR #199, opened 2019-03-25
Testing - stubs requests by webmock

ezr-ondrej:
I have recorded all the requests necessary to call simple `get_cluster`.
Not sure if this is the right way to test, but sure it was fun exercise :laughing:
Request count: `16` of what `5` completely redundant.

I believe this is not that bad way to go, even though it requires a lot of data files, it would give us some relevant integration testing and we could actually find performance issues easily as we can easily find out how many times are given stubs called.

@timogoebel any thoughts?


timogoebel:
> I believe this is not that bad way to go, even though it requires a lot of data files, it would give us some relevant integration testing and we could actually find performance issues easily as we can easily find out how many times are given stubs called.

I fully agree. This is a lot of work, but a really good thing.

chris1984:
@ezr-ondrej is this ready to go?

ezr-ondrej:
Closing in favor of #209

PR #202, opened 2019-03-26
Fixes #201 - added disks on vm cloning

### LINKED ISSUE #201: Create storagePod disks correctly while cloning
As of now, adding more disk to the storagePod will fail on cloning the VM.

ezr-ondrej:
If more disks are specified and their placement is on storagePod, cloning fails.
This is a partial fix though, because now if storagePod is specified, it places all the disks on that pod, no matter what the disks have specified.
It will clone the machine successfuly though.

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -164,7 +164,7 @@ module Fog
             device_change << modify_template_nics_simple_spec(options['network_label'], options['nic_type'], options['network_adapter_device_key'], options['datacenter'])
           end
           if disks = options['volumes']
-            device_change.concat(modify_template_volumes_specs(vm_mob_ref, options['volumes']))
+            device_change.concat(modify_template_volumes_specs(vm_mob_ref, options['volumes'], default_storage_pod: options['storage_pod']))
           end
           virtual_machine_config_spec.deviceChange = device_change if device_change.any?
           # Options['numCPUs'] or Options['memoryMB']
@@ -789,7 +789,7 @@ module Fog
           specs
         end
 
-        def modify_template_volumes_specs(vm_mob_ref, volumes)
+        def modify_template_volumes_specs(vm_mob_ref, volumes, default_storage_pod: nil)
           template_volumes = vm_mob_ref.config.hardware.device.grep(RbVmomi::VIM::VirtualDisk)
           modified_volumes = volumes.take(template_volumes.size)
           new_volumes      = volumes.drop(template_volumes.size)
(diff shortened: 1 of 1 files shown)


chris1984:
This will also allow me to add another unit test to the image testing for regression tests ;)

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -789,14 +789,17 @@ def modify_template_nics_specs(vm_mob_ref, nics, datacenter)
           specs
         end
 
-        def modify_template_volumes_specs(vm_mob_ref, volumes)
+        def modify_template_volumes_specs(vm_mob_ref, volumes, default_storage_pod = nil)
```
timogoebel: can we make this a kwarg so it's easier to figure out what the method expects?

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -789,14 +789,17 @@ def modify_template_nics_specs(vm_mob_ref, nics, datacenter)
...
           template_volumes = vm_mob_ref.config.hardware.device.grep(RbVmomi::VIM::VirtualDisk)
           modified_volumes = volumes.take(template_volumes.size)
           new_volumes      = volumes.drop(template_volumes.size)
 
           specs = []
           template_volumes.zip(modified_volumes).each do |template_volume, new_volume|
             if new_volume
+              # copy identifiers fo fog device to mark them as used
```
timogoebel: typo: `to fog`

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -811,8 +814,7 @@ def modify_template_volumes_specs(vm_mob_ref, volumes)
                          device: template_volume }
             end
           end
-          new_volumes.map { |volume| volume.unit_number = volumes.index(volume) < 7 ? volumes.index(volume) : volumes.index(volume) + 1 }
```
timogoebel: This is solved by copying the identifiers from the template volume?

ezr-ondrej: Yeah exactly, the problem was the volume numbers were duplicated, but the above line is fixing it.

lib/fog/vsphere/requests/compute/vm_clone.rb
ezr-ondrej: :dart:

lib/fog/vsphere/requests/compute/vm_clone.rb
ezr-ondrej: yeah, feels better :+1:

chris1984:
Thanks @ezr-ondrej

PR #204, opened 2019-03-30
Fixes #203 - network lookup by its ref

### LINKED ISSUE #203: Allow network lookup by it's ref (what is effectively an id)

ezr-ondrej:
Make the lookup by ref available, what will effectively mean, that this is going to be valid:

```ruby
network_id = compute.networks.first.id # find my desired network

compute.servers.new interfaces: [{network: network_id}]

server = compute.servers.first
server.interfaces.first.network = network_id
```

And it is really convenient to work with id.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -178,7 +178,7 @@ module Fog
             )
 
           else
-            RbVmomi::VIM.VirtualEthernetCardNetworkBackingInfo(deviceName: nic.network)
+            RbVmomi::VIM.VirtualEthernetCardNetworkBackingInfo(deviceName: raw_network.name)
           end
         end
 
--- a/lib/fog/vsphere/requests/compute/get_network.rb
+++ b/lib/fog/vsphere/requests/compute/get_network.rb
@@ -2,16 +2,16 @@ module Fog
   module Vsphere
     class Compute
       class Real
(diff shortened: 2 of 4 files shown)


lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -5,6 +5,7 @@ class Real
         def create_vm(attributes = {})
           # build up vm configuration
 
+          pp attributes
```
timogoebel: I believe this debug code can safely be deleted.

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -58,6 +59,7 @@ def create_vm_on_storage_pod(storage_pod, vm_cfg, vmFolder, resource_pool, datac
             configSpec: vm_cfg,
             host: host
           )
+          pp storage_spec
```
timogoebel: This, too.

lib/fog/vsphere/requests/compute/create_vm.rb
ezr-ondrej: Forgot that one :disappointed: thanks :+1:

lib/fog/vsphere/requests/compute/create_vm.rb
ezr-ondrej: ^^ :dart:

PR #206, opened 2019-04-03
Bugfix/replace blank ref

loperaja:
the method blank? does not exist outside the rails world and throws errors. Changed it to make sure options['volumes'] is not nil and not an empty array


chris1984:
Thanks @loperaja

PR #207, opened 2019-04-16
Save cpu and memory HotAdd enabled from server model

scambra:
send cpuHotAddEnabled and memoryHotAddEnabled to ReconfigVM_Task spec, so it can be changed in server model


chris1984:
Thanks @scambra

Issue #208, opened 2019-04-17
Cover by meaningful test suit


ezr-ondrej:
Currently we do not have really meaningful test suit. It would be great to at least hit all the important methods from `Compute::Real` class.

PR #209, opened 2019-04-17
Refs #208 - add vcr webmock testing

ezr-ondrej:
POC of vcr cassette testing, as #199 was missing an order, which is brought by cassettes.

1. removes test files from gem build as the mock files are pretty huge and unnecessary to pack
2. Ads `vcr` dependency.
3. Creates an ability to record new vcrs in `recording_webmock_cassette` block
    - not fully there, still some hands on necessary,
    - helps to cover vsphere credentials though
    - do not record shared requests (logins for now)
4. Creates `shared.yml` cassette with the basic repeated requests in every example
5. Creates first few cassettes - most of get requests covered

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ tmp
 mkmf.log
 gemfiles/*.lock
 .*.swp
+tests/vsphere_config.yml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,55 @@
 # Getting Involved
 
-New contributors are always welcome, when it doubt please ask questions. We strive to be an open and welcoming community. Please be nice to one another.
+:+1::tada: First off, thanks for taking the time to contribute! :tada::+1:
+
+New contributors are always welcome, when it doubt please ask questions.
+We strive to be an open and welcoming community. Please be nice to one another.
+
+## Non-Coding
+
+- Offer feedback on open [issues](https://github.com/fog/fog-vsphere/issues).
+- Organize or volunteer at events.
 
 ## Coding
 
 - Pick a task:
-- Offer feedback on open [pull requests](https://github.com/fog/fog-vsphere/pulls).
 - Review open [issues](https://github.com/fog/fog-vsphere/issues) for things to help on.
 - [Create an issue](https://github.com/fog/fog-vsphere/issues/new) to start a discussion on additions or features.
 - Fork the project, add your changes and tests to cover them in a topic branch.
 - Commit your changes and rebase against `fog/fog-vsphere` to ensure everything is up to date.
 - [Submit a pull request](https://github.com/fog/fog-vsphere/compare/)
 
-## Non-Coding
+### Testing
 
-- Offer feedback on open [issues](https://github.com/fog/fog-vsphere/issues).
-- Organize or volunteer at events.
+We are in a process of a covering the functionality by integrational tests.
+We would like to cover all the requests made towards the vSphere instance.
+Doing so should give us more confidence and make the development easier.
+We will appreciate if you are able to help. Following steps will guide you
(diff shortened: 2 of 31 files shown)


ezr-ondrej:
@timogoebel @chris1984 thoughts? Basic argument against is that the test files are large, they could be deduplicated, but it should be accompanied with deduplicating the requests made to the server. What means work obviously :laughing:

timogoebel:
Nice, looks good. I'm wondering how we can allow everybody to (re-)capture the VCRs.

ares:
That's hard if we don't share some public vmware instance I think, but usually only some cassette needs updating, unless we do some big version bump. A readme entry for how to record new cassettes would be a good start.

ezr-ondrej:
> Nice, looks good. I'm wondering how we can allow everybody to (re-)capture the VCRs.

I have tried to do just that and I believe it is not far from achieving it. As @ares suggests, I would add an entry in README as a followup for making it easy to digest. But the prerequisite of Vsphere instance is there obviously, but I guess if I am contributing here, I got one on hands?

timogoebel:
> But the prerequisite of Vsphere instance is there obviously, but I guess if I am contributing here, I got one on hands?

Sure, but do I need a specific vsphere instance or can I just use a random one?

ezr-ondrej:
> > But the prerequisite of Vsphere instance is there obviously, but I guess if I am contributing here, I got one on hands?
> 
> Sure, but do I need a specific vsphere instance or can I just use a random one?

The idea was to have a cassettes per version, but I guess the best thing about vsphere API is that is pretty stable. At least as far as our coverage of it is concerned, so I wouldn't mind dropping that idea in favor of us being able to maintain the test suite easier. So what about putting the vsphere version as a cassette attribute and otherwise forget them for now?

ezr-ondrej:
> I wasn't thinking about the version. Let me put that differently: Can I record casettes without having access to your vSphere instance (read: just have access to my instance)?

yeah, that is possible through cassette erb variables:
Here you fill the info about your instance: https://github.com/fog/fog-vsphere/pull/209/files#diff-607791bf648f3efe993cb2473608b96dR14
And it is making requests against that instance and replaces the url of your instance before creating the cassette. 
https://github.com/fog/fog-vsphere/pull/209/files#diff-41503bac174f35f046ae72012f4a91fbR7

Your login details should not be written as those are send just to login and that is filtered out as it is in the `shared` cassette.

ezr-ondrej:
> I wasn't thinking about the version. Let me put that differently: Can I record casettes without having access to your vSphere instance (read: just have access to my instance)?

You still need to design the examples as it works with your instance though (right names of entities) and that is the hand work necessary if there is something you do not want to share.

And a DISCLAIMER, I just believe it should work, I did not tested it, in more complex scenario 🙂 Like putting a `find_datacenter` request in the shared examples. But it should still be achievable through the erb variables.

chris1984:
@ezr-ondrej @ares I have a public VMware instance we can use for testing? I bought it just for this so we can have tests that run on a PR etc.

ares:
Works for me, it would be great if @ezr-ondrej could try to record the same against this instance. We'll see how cassettes change. Hopefully it's not too much of work in order to setup the vmware data for all of these tests.

ezr-ondrej:
@chris1984 yeah, I would love to do that! at least I will reassure the cassette capturing is not instance dependent :)

chris1984:
@ezr-ondrej I will email you the info today so you have it and make you an admin account since it is public I disabled the Administrator account. @timogoebel did you want the info as well?

timogoebel:
> @timogoebel did you want the info as well?

Sure, that would be lovely.

ezr-ondrej:
I have added an documentation for the behaviour in CONTRIBUTING.md

PR #214, opened 2019-06-06
Grab DC if within a folder

chris1984:


lib/fog/vsphere/models/compute/datacenter.rb
```diff
@@ -24,7 +24,7 @@ def storage_pods(filters = {})
         end
 
         def vm_folders(filters = {})
-          service.folders({ datacenter: path.join('/'), type: :vm }.merge(filters))
+          service.folders({ datacenter: name, type: :vm }.merge(filters))
```
ezr-ondrej: Could we do it in all the methods, please? I guess we have the same bug in all of them.

chris1984:
@ezr-ondrej updated :)

chris1984:
fixing rubocop

ezr-ondrej:
As discussed offline, this doesn't work in all cases, we should make sure we are covering all the use cases first.

chris1984:
Getting this from Foreman:

http://rx-paste.usersys.redhat.com/view/1748dc7c

chris1984:
@ezr-ondrej can we merge this for now then work on the real issue later to close the bugzilla?

Tested it with the original code submit and it works fine:

Foreman host created:

https://nimb.ws/txXu1s

Host created in vCenter:

https://nimb.ws/eOmdT7

ezr-ondrej:
Ok, as this is fixing the issue in the folder and it doesn't break anything... The issues i had having are unrelated it seems and they appear even without the patch.

PR #231, opened 2019-08-15
Add Jenkinsfile for regression testing

chris1984:


chris1984:
This works correctly now, just need to work on the role/playbook to see why it is failing but the goal of this pr to make jenkins check prs is correct now and working. Merging.

PR #238, opened 2019-12-09
Fixes #237 - load simple clusters

### LINKED ISSUE #237: Cluster is not loaded for simple cluster
If the VM is not on the real cluster, the cluster is `nil`, but it is still a cluster, only simple one.

ezr-ondrej:
Simple clusters are `RbVmomi::VIM::ComputeResource` not `RbVmomi::VIM::ClusterComputeResource`

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -276,7 +276,7 @@ module Fog
                     when :datacenter
                       RbVmomi::VIM::Datacenter
                     when :cluster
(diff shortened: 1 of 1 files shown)


sbernhard:
@timogoebel do you agree with this change?

timogoebel:
Looks ok.

chris1984:
Merge away @ezr-ondrej

sbernhard:
This is a nasty issue, can you please create a new release and put it in foreman?

chris1984:
@sbernhard will make a new release

PR #249, opened 2020-03-25
fixes earlier nsx-t patches

fwdIT:
@timogoebel with this small fix the nsx-t network selection works again on the lastest commit on master of fog-vsphere making the earlier nsx-t patches active on the latest codebase
@ezr-ondrej if all is approved, I would appreciate a new package release and also a backport to the rpm repo of 1.24 if possible :)

thx all

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -29,7 +29,7 @@ module Fog
         # unitNumber: 7,
         #
         def list_vm_interfaces(vm_id, datacenter = nil)
-          get_raw_interfaces(vm_id, datacenter).map { |nic| raw_to_hash nic }
+          get_raw_interfaces(vm_id, datacenter).map { |nic| raw_to_hash(nic, datacenter) }
         end
 
         def get_vm_interface(vm_id, options = {})
@@ -61,7 +61,7 @@ module Fog
 
         private
 
-        def raw_to_hash(nic, datacenter = nil)
+        def raw_to_hash(nic, datacenter)
           if nic.backing.respond_to?(:network)
             network = nic.backing.network.name
           elsif nic.backing.respond_to?(:port)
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
```diff
@@ -61,7 +61,7 @@ def get_raw_interface(vm_id, options = {})
 
         private
 
-        def raw_to_hash(nic, datacenter)
+        def raw_to_hash(nic, datacenter = nil)
```
timogoebel: Thanks, is the method called here?

https://github.com/fog/fog-vsphere/blob/74fbbc9241b746c30ee64d2e54cd0310b8b6a6bf/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb#L32

I'm asking because a better fix would be to call the method with passing the datacenter and keeping this as a required parameter.

fwdIT: I started with a 100% clone of the master branch on fog-vsphere
Then did a diff with my working patched v3.2.1 and looked at the differences related to the nsx-t patches (files in which I did changes: list_networks.rb and list_vm_interfaces.rb)

The diff between my working v.3.2.1 and the master on fog-vsphere showed only differences in commented code:

`36a37
>           #raw_to_hash(raw) if raw
63a65,66
>         # rubocop:disable Style/ConditionalAssignment
>         #def raw_to_hash(nic)
68a72,74
>           #else
>           #  network = nil
>           #end`

timogoebel: Thanks, but I don't quite get what you're trying to tell me.

If we change the line I mentioned to:

```ruby
get_raw_interfaces(vm_id, datacenter).map { |nic| raw_to_hash(nic, datacenter) } 
```

Does that fix the issue on master as well?

fwdIT: sorry, was afk for a sec and realized in the meantime my reply was probably unclear, posting diffs for starters are quite silly because github shows diffs already :)
will try the suggested line while reverting the default nil value in the previous change

fwdIT: @timogoebel your patch worked :) hope this fix is more optimal

ezr-ondrej:
Thanks @fwdIT and @timogoebel.

PR #251, opened 2020-05-04
Fixed wrong variable type of eager zero / thin. They are "bool" and n…

sbernhard:
…ot "string"

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -300,7 +300,7 @@ module Fog
           file_operation = options[:file_operation] || (:create if operation == :add)
           payload[:fileOperation] = file_operation if file_operation
 
-          if operation == :add && disk.thin == 'false' && disk.eager_zero == 'true'
+          if operation == :add && disk.thin == false && disk.eager_zero == true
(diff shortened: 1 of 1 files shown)


sbernhard:
This is part 2 of https://projects.theforeman.org/issues/28874
Part 1 needs to be changed in foreman as eagerzero instead of eager_zero is sent. currently working on it.

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -300,7 +300,7 @@ def create_disk(disk, operation = :add, options = {})
           file_operation = options[:file_operation] || (:create if operation == :add)
           payload[:fileOperation] = file_operation if file_operation
 
-          if operation == :add && disk.thin == 'false' && disk.eager_zero == 'true'
+          if operation == :add && disk.thin == false && disk.eager_zero == true
```
timogoebel, line 303: I know, I'm late to the party, but could you have just used `!disk.thin && disk.eager_zero` instead?

sbernhard: Fixed with https://github.com/theforeman/foreman/pull/7634

PR #252, opened 2020-05-04
Eager zero fixes

sbernhard:


chris1984:
@sbernhard will start to test on Satellite and Foreman

sbernhard:
still not done. eager_zero will now be returned from fog to foreman but foreman expects eagerzero and fog sends eager_zero :(

sbernhard:
how can I fix these 2 codeclimate issues?

timogoebel:
> how can I fix these 2 codeclimate issues?

Definitely not in the scope of this PR.

lib/fog/vsphere/requests/compute/list_vm_volumes.rb
```diff
@@ -28,6 +28,8 @@ class Real
         #  unitNumber: 1
         # )]
 
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/MethodLength
```
ezr-ondrej, line 32: @sbernhard could you pleaser remove these? It's not necessary because of config in `.rubocop.yml`, one is disabled and one is set on very high number, so we don't need to disable it localy. That should enable the tests to at least run :)

ezr-ondrej: The codeclimate will just complain anyway, but we have so many issues in this repo, that its not real problem.

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -838,6 +838,7 @@ def modify_template_volumes_specs(vm_mob_ref, volumes)
               end
               template_volume.backing.diskMode = new_volume.mode
               template_volume.backing.thinProvisioned = new_volume.thin
+              template_volume.backing.eagerlyScrub = new_volume.eager_zero
```
ezr-ondrej, line 841: Oh, we cannot have `thin` and `eager_zero` at the same time. Last thing then, sorry for not noticing before:
```suggestion
              template_volume.backing.eagerlyScrub = !new_volume.thin && new_volume.eager_zero
```

sbernhard:
Ah, right. It would be good that the UI (= foreman) validates, that a disk can not be set to thin and eager_zero together.

ezr-ondrej:
> Ah, right. It would be good that the UI (= foreman) validates, that a disk can not be set to thin and eager_zero together.

Will do in the takeover PR :+1:

ezr-ondrej:
I've get that safeguard in the foreman PR, can we get that in this PR anyway so it's safe for other users as well?

ezr-ondrej:
Thanks @sbernhard !

Issue #256, opened 2020-07-14
list_resource_pools returns incorrect resource pool name 


rlazoryshchak:
When folder_path(resource_pool) for example '/Datacenters/DC1/host/vc-qa-DC1-cluster1/Resources/RP1/Resources'
Then we get unexpected resource pool name:
```
[2] pry(main)> path='/Datacenters/DC1/host/vc-qa-DC1-cluster1/Resources/RP1/Resources'
=> "/Datacenters/DC1/host/vc-qa-DC1-cluster1/Resources/RP1/Resources"
[3] pry(main)> path.gsub(/^.*Resources(\/|)/, '')
=> ""
[4] pry(main)> path.sub(/^.*Resources(\/|)/, '')
=> ""
```
But should be `"RP1/Resources"`

rlazoryshchak:
@ezr-ondrej or @chris1984 could someone of you check this:

fog-vsphere/lib/fog/vsphere/requests/compute/list_resource_pools.rb:39
```
name = folder_path(resource_pool).gsub(/^.*Resources(\/|)/, '')
name = 'Resources' if name.empty?
```
could be replaced by:
```
path = folder_path(resource_pool)
name = /(?<=Resources\/)(.+)/.match(path) ? $1 : 'Resources'
```

PR #257, opened 2020-07-15
list_resource_pools returns incorrect resource pool name #256

rlazoryshchak:
When some resource pool('ParentResourcePool') has nested resource pool('Resources') we get this child resource pool with name 'Resources' without path of parent resource pool('Resources' instead of 'ParentResourcePool/Resources').


rlazoryshchak:
@chris1984, @ezr-ondrej when you have a moment, please take a look.

Issue #261, opened 2020-07-30
Fix resource pool creation with '/' in name


rlazoryshchak:
```
compute.create_resource_pool(
	datacenter: 'DC1', 
	cluster: 'vc-qa-DC1-cluster1', 
	name: 'Resource/Pool1',
)
```
This request will failed such like '/' character present in name.

PR #262, opened 2020-07-30
Fix resource pool creation with '/' in name #261

rlazoryshchak:
Now we can create resource pool with '/' in resource pool and root resource pool namespace.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_resource_pool.rb
+++ b/lib/fog/vsphere/requests/compute/create_resource_pool.rb
@@ -6,17 +6,19 @@ module Fog
           cluster = get_raw_cluster(attributes[:cluster], attributes[:datacenter])
 
           root_resource_pool = if attributes[:root_resource_pool_name]
-                                 cluster.resourcePool.find attributes[:root_resource_pool_name]
+                                 cluster.resourcePool.find attributes[:root_resource_pool_name].gsub('/', '%2f')
                                else
                                  cluster.resourcePool
                                end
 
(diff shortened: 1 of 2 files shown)


rlazoryshchak:
@ezr-ondrej and @chris1984 could you take a look?

lib/fog/vsphere/requests/compute/create_resource_pool.rb
```diff
@@ -6,17 +6,19 @@ def create_resource_pool(attributes = {})
           cluster = get_raw_cluster(attributes[:cluster], attributes[:datacenter])
 
           root_resource_pool = if attributes[:root_resource_pool_name]
-                                 cluster.resourcePool.find attributes[:root_resource_pool_name]
+                                 cluster.resourcePool.find attributes[:root_resource_pool_name].gsub('/', '%2f')
```
ezr-ondrej, line 9: I guess this is the main part of the fix?
Why we do not need this in another resources?
I thought that translating this characters would be a job for the underlying communication library :thinking:

ezr-ondrej:
@rlazoryshchak could you add a comment why do we need the transpolation of `/`, please?
Apart of it :+1:

rlazoryshchak:
This is fix naming of resource pools which contains '/' symbol. Bug coused by rbvmomi transpolation '/' to '%2f' and while creating nested resource pool we can't find it in parent resource pool which contains '/'.

ezr-ondrej:
I've added the comment in the commit comment to be easily found in the future, thanks @rlazoryshchak :+1:

PR #263, opened 2020-10-06
clone vm in different datacenter

scambra:
if datacenter option is in clone arguments, add vm's datacenter as template_datacenter, as vm to use as template may be in different datacenter

### DIFF
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -157,7 +157,11 @@ module Fog
 
           # Give our path to the request
           req_options['template_path'] = "#{relative_path}/#{name}"
-          req_options['datacenter'] = datacenter.to_s
+          if req_options['datacenter']
+            req_options['template_datacenter'] = datacenter.to_s
+          else
+            req_options['datacenter'] = datacenter.to_s
+          end
 
           # Perform the actual clone
           clone_results = service.vm_clone(req_options)
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -157,11 +157,8 @@ module Fog
 
           # Give our path to the request
           req_options['template_path'] = "#{relative_path}/#{name}"
-          if req_options['datacenter']
-            req_options['template_datacenter'] = datacenter.to_s
-          else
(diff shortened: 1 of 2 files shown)


lib/fog/vsphere/models/compute/server.rb
```diff
@@ -157,7 +157,11 @@ def clone(options = {})
 
           # Give our path to the request
           req_options['template_path'] = "#{relative_path}/#{name}"
-          req_options['datacenter'] = datacenter.to_s
+          if req_options['datacenter']
+            req_options['template_datacenter'] = datacenter.to_s
+          else
+            req_options['datacenter'] = datacenter.to_s
+          end
```
ezr-ondrej, line 164: I think, we always want to set the `template_datacenter`, right? :thinking:
What about this:
```suggestion
          req_options['template_datacenter'] = datacenter.to_s
          req_options['datacenter'] ||= datacenter.to_s
```

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -104,14 +110,11 @@ def vm_clone(options = {})
           # Option handling
           options = vm_clone_check_options(options)
 
-          # Added for people still using options['path']
-          template_path = options['path'] || options['template_path']
```
ezr-ondrej, line 108: :+1: I bet there is more dead code like this xD

ezr-ondrej (commented):
Great addition @scambra ! :+1:
Just one nit, let me know what you think :)
Love the dead code deletion! please more of that xD

lib/fog/vsphere/models/compute/server.rb
scambra: I can change that, I thought it wouldn't be necessary because vm_clone uses options['template_datacenter'] || options['datacenter'], so template_datacenter was needed only if datacenter may be different, but it may be better to set template_datacenter explicitly

ezr-ondrej (approved):
Thanks @scambra ! :+1:

ezr-ondrej:
Do you need this released fast or can it wait a while?

chris1984:
Starting to kick off regression tests

scambra:
I can wait, I can use repository in gemfile until it's released

Issue #265, opened 2020-10-27
Expose _ref attribute of Compute::Network


pendor:
The `get_network` and `get_raw_network` functions in Fog::Vsphere::Compute::Real [lib/fog/vsphere/requests/compute/get_network.rb](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/requests/compute/get_network.rb) accept either a Name or a _ref attribute of a network object.  The Fog::Vsphere::Compute::Network object in [lib/fog/vsphere/models/compute/network.rb](https://github.com/fog/fog-vsphere/blob/master/lib/fog/vsphere/models/compute/network.rb) doesn't currently expose the _ref attribute as a publicly accessible value, even though the data is present in the underlying ActiveRecord from the API call to VSphere.  

The _ref attribute should be exposed so that it can be used to later call `get_network` with a globally unique value.  Name may not be unique in some environments.  Additionally, at least one client of fog-vsphere (Foreman) currently incorrectly uses the visible id attribute in place of the _ref attribute when calling `get_raw_network`.  This works in many but not all instances as id is often but not always the same value as _ref, depending on how VSphere objects have been edited over time.

ezr-ondrej:
Thanks a lot @pendor! This has been released as part of v3.5.0 release.

PR #264, opened 2020-10-14
Unify network ID attribute

ezr-ondrej:
Use `_ref` as ID for all kinds of networks.

See https://community.theforeman.org/t/could-not-match-network-interface-vmware-again/20862 for the detailed problem description.

PR #266, opened 2020-10-27
Expose _ref from Compute::Network #265

pendor:
The get_network function in lib/fog/vsphere/requests/compute/get_network.rb
takes a name or _ref attribute.  This commit expose that _reg attribute
(already retrieved via previous vsphere call) so that clients of
fog-vsphere can look up network objects via the globally unique _ref
attribute instead of the non-unique Name attribute.

Additionally, this enables fixing a bug in clients where the id attribute
is incorrectly substituted for _ref, retrieving an incorrect object in
certain VSphere environments.

### DIFF
--- a/lib/fog/vsphere/models/compute/network.rb
+++ b/lib/fog/vsphere/models/compute/network.rb
@@ -9,6 +9,7 @@ module Fog
         attribute :accessible # reachable by at least one hypervisor
         attribute :virtualswitch
         attribute :vlanid
+        attribute :_ref
 
         def to_s
           name


pendor:
Would it be possible to have the test build re-run for this and for someone to take a look at merging it?  As far as I can tell, but build failed for reasons unrelated to this change as the preceding builds also failed with the same issue.

chris1984:
@pendor I will take a look at this, test and merge

ezr-ondrej:
Eh, it was just a RuboCop complaining, we should improve our rubocop notifications. I've fixed that.

ezr-ondrej (approved):
I don't love this solution (`_ref` is quite internal attribute), but the alternative is: https://github.com/fog/fog-vsphere/pull/264.
I have no way of testing the other one, so this seems safer and we already basically expose `_ref` by implying that we are expecting it in `get_network`, so I'm ok with this.

Issue #269, opened 2021-07-14
Snapshot of running VM always includes memory


m-bucher:
https://github.com/fog/fog-vsphere/blob/6333944345ae0dd036e2605709ec06ca317c3441/lib/fog/vsphere/requests/compute/vm_take_snapshot.rb#L12
This came up in https://github.com/ATIX-AG/foreman_snapshot_management/issues/67

Also: is there a reason `quiesce` is not set to true by default? From [the description in the vmware-docs](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/b525fb12-61bb-4ede-b9e3-c4a1f8171510/99ba073a-60e9-4933-8690-149860ce8754/doc/vim.VirtualMachine.html#createSnapshot) it sounds like something that one would like to have enabled.

chris1984:
will start to test

PR #270, opened 2021-07-14
Fix omitted snapshot-option

m-bucher:
Solves #269

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_take_snapshot.rb
+++ b/lib/fog/vsphere/requests/compute/vm_take_snapshot.rb
@@ -5,13 +5,17 @@ module Fog
         def vm_take_snapshot(options = {})
           raise ArgumentError, 'instance_uuid is a required parameter' unless options.key? 'instance_uuid'
           raise ArgumentError, 'name is a required parameter' unless options.key? 'name'
+          defaults = {
+            description: '',
+            memory: true,
+            quiesce: false
+          }
+          opts = options.clone
+          defaults.each do |k, v|
+            opts[k] = v unless opts.key?(k) || opts.key?(k.to_s)
+          end
           vm = get_vm_ref(options['instance_uuid'])
-          task = vm.CreateSnapshot_Task(
-            name: options['name'],
-            description: options['description'] || '',
-            memory: options['memory'] || true,
-            quiesce: options['quiesce'] || false
-          )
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/vm_take_snapshot.rb
```diff
@@ -5,13 +5,13 @@ class Real
...
+            quiesce: false
+          }
           vm = get_vm_ref(options['instance_uuid'])
-          task = vm.CreateSnapshot_Task(
-            name: options['name'],
-            description: options['description'] || '',
-            memory: options['memory'] || true,
-            quiesce: options['quiesce'] || false
-          )
+          task = vm.CreateSnapshot_Task(defaults.merge(options))
```
ezr-ondrej, line 14: this won't work if I'm correct, options have (or can have at least) string keys.
```suggestion
          task = vm.CreateSnapshot_Task(defaults.merge(options.transform_keys(&:to_sym)))
```
This requires ruby `>= 2.5` but that should not be an issue I believe :thinking:

ezr-ondrej: or just have defaults in strings and continue expecting options will have string keys :)

m-bucher: The gemspec wants ruby `>=2.0.0`

lib/fog/vsphere/requests/compute/vm_take_snapshot.rb
```diff
@@ -5,13 +5,17 @@ class Real
...
           raise ArgumentError, 'name is a required parameter' unless options.key? 'name'
+          defaults = {
+            description: '',
+            memory: true,
+            quiesce: false
+          }
+          opts = options.clone
+          defaults.each do |k, v|
+            opts[k] = v unless opts.key?(k) || opts.key?(k.to_s)
+          end
```
m-bucher, line 16: Not as smooth, but works with ruby-2.0 as well :wink:

ezr-ondrej:
@chris1984 could you take it for a spin?

ezr-ondrej (approved):
LGTM :+1:

chris1984:
Yeah will do

chris1984 (approved):
ACK, works good and no regressions with Foreman

PR #271, opened 2021-11-11
Start nic connected

ezr-ondrej:


ezr-ondrej:
It was not the issue in the first place

PR #277, opened 2022-06-29
Fixes #35125 - Avoid unneeded "get_raw_datacenter" call

hao-yu:
Use both name and path to match datacenter from the local cache
which will avoid fetching from Vcenter again.


chris1984:
@Ron-Lavi will test this one too, while i'm here and get a new gem out

chris1984:
ACK, tested and works well without any regressions.

PR #281, opened 2022-11-18
Switch to rbvmomi2

ekohl:
rbvmomi2 is a maintained fork of rbvmomi.

I'll admit this is completely untested.


chris1984:
I will test this on my devel box to see if it does not cause regression with the switch

sbernhard:
Thanks to bring this up. I guess, it will be pretty good if we share the same lib with ManageIQ.

PR #292, opened 2024-02-15
Fixes vmware 8 u2 iso detach

Manisha15:


chris1984:
Will start to test and get it added to our board

sbernhard:
> Will start to test and get it added to our board

To fix the vmware 8 issue, you need to have https://github.com/theforeman/foreman_bootdisk/pull/158, too

lib/fog/vsphere/requests/compute/get_vm_first_sata_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Fog
+  module Vsphere
+    class Compute
+      class Real
+        def get_vm_first_sata_controller(vm_id)
+          ctrl = get_vm_ref(vm_id).config.hardware.device.find { |device| device.is_a?(RbVmomi::VIM::VirtualAHCIController) }
```
ekohl, line 6: Just reading over this: can't this return `nil`? If so, the rest of the code would crash.

chris1984: +1 to using ruby safe navigation operator. I have not looked at the code yet, I had to redeploy vcenter 8, so will test this this week. But yeah this needs a fix

ekohl: Don't you just want to return `nil` if it's not found instead?

Manisha15: Updated the code to raise `Not Found ` error if it's nil based on other compute/requests.

chris1984:
Thanks! Will get this tested today/tomorrow then look at the code again

chris1984 (approved):
Code looks fine, thanks for addressing Ewoud comments. Test failures are unrelated, have a card on the board to get them fixed.

Work with PR and bootdisk_pr

Before:

Traceback from `production.log`
```ruby
==> /var/log/foreman/production.log <==
2024-03-04T14:37:50 [W|app|ed510f37] Failed to detach ISO image from CDROM drive of instance sean-horth.lab.eng.rdu2.redhat.com: InvalidPowerState: The attempted operation cannot be performed in the current state (Powered on).
2024-03-04T14:37:50 [I|app|ed510f37] Backtrace for 'Failed to detach ISO image from CDROM drive of instance sean-horth.lab.eng.rdu2.redhat.com: InvalidPowerState: The attempted operation cannot be performed in the current state (Powered on).' error (RbVmomi::Fault): InvalidPowerState: The attempted operation cannot be performed in the current state (Powered on).
 ed510f37 | /usr/share/gems/gems/rbvmomi2-3.7.0/lib/rbvmomi/vim/Task.rb:15:in `wait_for_completion'
```

Screenshot showing error in VMware:

before

With PR, no traceback in production log and we can see good tasks in VMware:

after
after1

PR #294, opened 2024-03-26
Add an exception for the missing storage field

girijaasoni:
This is a proposed solution for [#2186114](https://bugzilla.redhat.com/show_bug.cgi?id=2186114)
I think creating a host without mentioning the storage pod or the datastore should not be permitted. Adding validation is one way to go but achieving that is extremely difficult. Adding an exception is a much simpler approach to it.

I'm open for any other ideas to approach this problem. Please feel free to share your thoughts on this.
cc: @shimshtein


ShimShtein:
A bit of context: currently the system does not allow empty storage field, and fails the creation of a VM with "Invalid path" error, which is not indicative at all.

chris1984 (commented):
Code wise it looks fine, will start testing

ShimShtein:
LGTM code-wise.

PR #54, opened 2016-10-18
vms can have more than one scsi controller

timogoebel:
This commit adds the ability to add more than one scsi controller to a vm.

This can be tested with:

```ruby
attrs = {
  ...
  :volumes => [
    {
      :storage_pod=>"Test",
      :name=>"Hard disk 1",
      :size_gb=>"10",
      :thin=>"true",
      :eager_zero=>"false",
      :controller_key => 1000
    },
    {
      :storage_pod=>"Test",
      :name=>"Hard disk 2",
      :size_gb=>"10",
      :thin=>"true",
      :eager_zero=>"false",
      :controller_key => 1001
    }
  ],
  :scsi_controllers => [
    {
      :type=>"VirtualLsiLogicController",
      :key => 1000
    },
    {
      :type=>"VirtualLsiLogicController",
      :key => 1001
    }
  ]
}

vm = client.servers.new(attrs)
vm.save
```

PR #300, opened 2024-05-15
Add support for NVME Controllers

girijaasoni:
This is the solution for the [feature](https://bugzilla.redhat.com/show_bug.cgi?id=2156673) request. 
NVME (Non Volatile Memory Express) controllers are a class above the currently supported (SCSI Controllers) and has multiple [advantages](https://www.techtarget.com/searchstorage/definition/NVMe-non-volatile-memory-express)

Currently Foreman and fog-vsphere are designed to only support SCSI controllers and it's types. The goal is to extend it to adding NVME and in future, other types. Do give this a read [SCSI, SATA, and NVMe Storage Controller Conditions, Limitations, and Compatibility](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.vm_admin.doc/GUID-5872D173-A076-42FE-8D0B-9DB0EB0E7362.html) 

Front End PR : https://github.com/theforeman/foreman/pull/10168
Hammer doc PR : https://github.com/theforeman/hammer-cli-foreman/pull/628

More details: https://docs.google.com/document/d/1hs4JekeCQP6brbZGg8hS8kcHohselH1HTEEhThgw-AE/edit#heading=h.lrnpszeygrmo

Ready for testing :D

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -43,6 +43,7 @@ module Fog
       model :customfield
       collection :customfields
       model :scsicontroller
+      model :nvmecontroller
       model :process
       model :cdrom
       collection :cdroms
@@ -111,6 +112,7 @@ module Fog
       request :list_customfields
       request :get_vm_first_scsi_controller
       request :list_vm_scsi_controllers
+      request :list_vm_nvme_controllers
       request :set_vm_customvalue
       request :vm_take_snapshot
       request :list_vm_snapshots
@@ -136,6 +138,7 @@ module Fog
       request :host_start_maintenance
       request :host_finish_maintenance
       request :get_vm_first_sata_controller
+      request :get_vm_first_nvme_controller
 
       module Shared
         attr_reader :vsphere_is_vcenter
@@ -346,10 +349,12 @@ module Fog
         end
       end
 
+      # rubocop:disable Metrics/ClassLength
       class Mock
         include Shared
         # rubocop:disable Metrics/MethodLength
         def self.data
+          # rubocop:disable Metrics/BlockLength
           @data ||= Hash.new do |hash, key|
             hash[key] = {
               servers: {
@@ -380,6 +385,11 @@ module Fog
                        'type'        => 'VirtualLsiLogicController',
                        'unit_number' => 7,
                        'key'         => 1000 }],
+                  'nvme_controllers' =>
+                       [{
+                         'type'        => 'VirtualNVMEController',
+                          'key'         => 2000
+                       }],
                   'interfaces'       =>
                     [{ 'mac' => '00:50:56:a9:00:28',
                        'network' => 'dvportgroup-123456',
@@ -616,6 +626,7 @@ module Fog
               }
             }
           end
+          # rubocop:enable Metrics/BlockLength
         end
 
         # rubocop:enable Metrics/MethodLength
@@ -637,6 +648,7 @@ module Fog
           self.class.data.delete(@vsphere_username)
         end
       end
+      # rubocop:enable Metrics/ClassLength
 
       class Real
         include Shared
--- a/lib/fog/vsphere/models/compute/server.rb
+++ b/lib/fog/vsphere/models/compute/server.rb
@@ -3,7 +3,7 @@ require 'fog/compute/models/server'
 module Fog
   module Vsphere
     class Compute
-      class Server < Fog::Compute::Server
+      class Server < Fog::Compute::Server # rubocop:disable Metrics/ClassLength
         extend Fog::Deprecation
         deprecate(:ipaddress, :public_ip_address)
         deprecate(:scsi_controller, :scsi_controllers)
@@ -49,6 +49,7 @@ module Fog
         attribute :guest_id
         attribute :hardware_version
         attribute :scsi_controllers, type: :array
+        attribute :nvme_controllers, type: :array
         attribute :cpuHotAddEnabled
         attribute :memoryHotAddEnabled
         attribute :firmware
@@ -60,9 +61,10 @@ module Fog
           super defaults.merge(attributes)
           self.instance_uuid ||= id # TODO: remvoe instance_uuid as it can be replaced with simple id
           initialize_interfaces
-          initialize_volumes
           initialize_customvalues
           initialize_scsi_controllers
+          initialize_nvme_controllers
+          initialize_volumes
         end
 
         # Lazy Loaded Attributes
@@ -289,6 +291,10 @@ module Fog
           attributes[:scsi_controllers] ||= service.list_vm_scsi_controllers(id)
         end
 
+        def nvme_controllers
+          attributes[:nvme_controllers] ||= service.list_vm_nvme_controllers(id)
+        end
+
         def scsi_controller
           scsi_controllers.first
         end
@@ -348,16 +354,21 @@ module Fog
           end
         end
 
+        def unassigned_volumes?
+          attributes[:volumes]&.any? { |vol| !vol.key?(:controller_key) } || false
+        end
+
+        def update_controller_key(vol)
+          vol.controller_key ||= attributes[:scsi_controllers].first&.key || 1000
+        end
+
         def initialize_volumes
-          if attributes[:volumes] && attributes[:volumes].is_a?(Array)
-            attributes[:volumes].map! do |vol|
-              if vol.is_a?(Hash)
-                service.volumes.new({ server: self }.merge(vol))
-              else
-                vol.server = self
-                vol
-              end
-            end
+          return unless attributes[:volumes].is_a?(Array)
+          attributes[:volumes].map! do |vol|
+            vol = service.volumes.new({ server: self }.merge(vol)) if vol.is_a?(Hash)
+            vol.server = self
+            update_controller_key(vol)
+            vol
           end
         end
 
@@ -368,19 +379,23 @@ module Fog
         end
 
         def initialize_scsi_controllers
-          if attributes[:scsi_controllers] && attributes[:scsi_controllers].is_a?(Array)
+          if attributes[:scsi_controllers].is_a?(Array) && !attributes[:scsi_controllers].empty?
             attributes[:scsi_controllers].map! do |controller|
               controller.is_a?(Hash) ? Fog::Vsphere::Compute::SCSIController.new(controller) : controller
             end
-          elsif attributes[:scsi_controller] && attributes[:scsi_controller].is_a?(Hash)
+          elsif attributes[:scsi_controller].is_a?(Hash) && !attributes[:scsi_controller].empty?
             attributes[:scsi_controllers] = [
               Fog::Vsphere::Compute::SCSIController.new(attributes[:scsi_controller])
             ]
-          elsif attributes[:volumes] && attributes[:volumes].is_a?(Array) && !attributes[:volumes].empty?
-            # Create a default scsi controller if there are any disks but no controller defined
-            attributes[:scsi_controllers] = [
-              Fog::Vsphere::Compute::SCSIController.new
-            ]
+          end
+          attributes[:scsi_controllers] = [Fog::Vsphere::Compute::SCSIController.new] if !attributes[:scsi_controllers]&.any? && unassigned_volumes?
+        end
+
+        def initialize_nvme_controllers
+          if attributes[:nvme_controllers].is_a?(Array)
+            attributes[:nvme_controllers].map! do |controller|
+              controller.is_a?(Hash) ? Fog::Vsphere::Compute::NVMEController.new(controller) : controller
+            end
           end
         end
       end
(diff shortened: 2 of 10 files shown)


lib/fog/vsphere/models/compute/server.rb
```diff
@@ -348,9 +359,25 @@ def initialize_interfaces
           end
         end
 
+        def unassigned_volumes?
+          attributes[:volumes]&.map! do |vol|
+            return true unless vol.key?(:controller_key)
+          end
+          false
```
ekohl, line 366: I wonder why you're using `map!`. Wouldn't that potentially wipe data? Isn't this what you're looking for?
```suggestion
          attributes[:volumes]&.any? { |vol| !vol.key?(:controller_key) } || false
```

Technically `|| false` isn't needed since `nil` and `false` evaluate to the same, but it ensures a proper boolean is returned.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -289,10 +292,18 @@ def scsi_controllers
...
+          attributes[:nvme_controllers] ||= service.list_vm_nvme_controllers(id)
+        end
+
         def scsi_controller
           scsi_controllers.first
         end
 
+        def nvme_controller
+          nvme_controllers.first
+        end
```
ekohl, line 305: From reading the above you're adding this, but deprecating it in the same PR. Wouldn't it be better to not introduce it at all?

ekohl (commented):
For RuboCop it's probably best to disable it in that class.

lib/fog/vsphere/models/compute/server.rb
girijaasoni: I suppose we can just not deprecate it. Scsi controllers were deprecated in https://github.com/fog/fog-vsphere/pull/54/commits/60a5dc320973ff6425bc941f24bae4e9187b57fa#diff-49106de6ade06895f1a509db7c9cd040ed76bfdb472548d8b99efc0468dce548R9 , not sure why
I tried to mirror the behavior but I guess we can undo that. @chris1984 , any idea why scsi_controllers were deprecated

lib/fog/vsphere/models/compute/server.rb
girijaasoni: that's a much better approach, Thanks!

lib/fog/vsphere/models/compute/server.rb
ekohl: Note it's the singular `scsi_controller` that's deprecated. IMHO it makes sense if you can have more than one that there isn't a singular version. Since it was done in 2016, by this point you can probably delete it.

lib/fog/vsphere/models/compute/server.rb
ekohl: I recommend reading https://docs.ruby-lang.org/en/master/Enumerable.html because it's extremely useful to know. Many things include `Enumerable`, making your life easier.

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -368,18 +395,28 @@ def initialize_customvalues
...
+            attributes[:scsi_controllers] = [Fog::Vsphere::Compute::SCSIController.new]
+          end
+        end
+
+        def initialize_nvme_controllers
+          if attributes[:nvme_controllers] && attributes[:nvme_controllers].is_a?(Array) && !attributes[:nvme_controllers].empty?
+            attributes[:nvme_controllers].map! do |controller|
+              controller.is_a?(Hash) ? Fog::Vsphere::Compute::NVMEController.new(controller) : controller
+            end
+          elsif attributes[:nvme_controller] && attributes[:nvme_controller].is_a?(Hash)
```
ShimShtein, line 417: Since we decided that `nvme_controller` should be removed, this part will need to be removed also.

ShimShtein (changes requested):
We also need to add mocked data to https://github.com/fog/fog-vsphere/blob/4a7f4e7f96b41190a5b6733112a84f8af8aabb6d/lib/fog/vsphere/compute.rb#L352 to contain at least one host with NVME controller and a drive that is attached to it.

ShimShtein:
Not tested, but code-wise looks good

chris1984:
Starting to test

lib/fog/vsphere/requests/compute/get_vm_first_nvme_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Fog
+  module Vsphere
+    class Compute
+      class Real
+        def get_vm_first_nvme_controller(vm_id)
+          # require 'pry'
+          # binding.pry
+          ctrl = get_vm_ref(vm_id).config.hardware.device.find { |device| device.is_a?(RbVmomi::VIM::VirtualNVMEController) }
+          raise(Fog::Vsphere::Compute::NotFound) unless ctrl
```
ekohl, line 9: I'd be explicit what wasn't found
```suggestion
          raise Fog::Vsphere::Compute::NotFound, "No NVME controller found for #{vm_id}" unless ctrl
```

lib/fog/vsphere/requests/compute/get_vm_first_nvme_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Fog
+  module Vsphere
+    class Compute
+      class Real
+        def get_vm_first_nvme_controller(vm_id)
+          # require 'pry'
+          # binding.pry
```
ekohl, line 7:
```suggestion
```

girijaasoni: ready for re-review

lib/fog/vsphere/models/compute/server.rb
```diff
@@ -348,15 +354,28 @@ def initialize_interfaces
...
-                vol
-              end
+          return unless attributes[:volumes].is_a?(Array)
+          attributes[:volumes].map! do |vol|
+            update_controller_key(vol)
+            if vol.is_a?(Hash)
+              service.volumes.new({ server: self }.merge(vol))
+            else
+              vol.server = self
+              vol
```
ShimShtein, line 378: If you reorder the calls to:
```suggestion
            if vol.is_a?(Hash)
              vol = service.volumes.new({ server: self }.merge(vol))
            end
            vol.server = self
            update_controller_key(vol)         
            vol
```

You will be able to remove the `case` statement in `update_controller_key`:
``` ruby
        def update_controller_key(vol)
            vol.controller_key ||= attributes[:scsi_controllers].first&.key || 1000
        end
```

because `vol` will always contain an object.

lib/fog/vsphere/requests/compute/get_vm_first_nvme_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+          {
+            type: ctrl&.class.to_s,
+            device_info: ctrl&.deviceInfo,
+            bus_number: ctrl&.busNumber,
+            key: ctrl&.key
+          }
+        end
+      end
+      class Mock
+        def get_vm_first_nvme_controller(vm_id); end
```
ShimShtein, line 17: The mock class returns `nil`, while the real method will throw an exception and never return `nil`. 
I think it would be nice to have the behavior be equal

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -887,7 +886,6 @@ def relocation_volume_backing(volume)
         end
       end
 
-      # rubocop:enable Metrics/ClassLength
```
ShimShtein, line 890: I think you should reenable the class length metric after the class definition.

ShimShtein (changes requested):

lib/fog/vsphere/compute.rb
```diff
@@ -346,11 +349,11 @@ def is_uuid?(id)
         end
       end
 
-      class Mock
+      class Mock # rubocop:disable Metrics/ClassLength
         include Shared
         # rubocop:disable Metrics/MethodLength
         def self.data
-          @data ||= Hash.new do |hash, key|
+          @data ||= Hash.new do |hash, key| # rubocop:disable Metrics/BlockLength
```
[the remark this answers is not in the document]
girijaasoni: Screenshot from 2024-06-24 16-14-27
I think its not required anymore

lib/fog/vsphere/models/compute/server.rb
girijaasoni: Interesting. Thanks Shim, updated :)

chris1984:
Starting to test again, @ekohl have all your concerns been addressed? If so will merge and get a release out pending all testing works again with the latest pushes.

lib/fog/vsphere/compute.rb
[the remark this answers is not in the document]
ekohl: It's the difference between an inline comment and one that is on its own.

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -255,7 +263,7 @@ def create_controller(controller = nil, index = 0)
             operation: options[:operation],
             device: controller_class.new(key: options[:key] || (1000 + index),
                                          busNumber: options[:bus_id] || index,
-                                         **(options[:type] == RbVmomi::VIM::VirtualAHCIController ? {} : {sharedBus: controller_get_shared_from_options(options)}))
+            **device_options)
```
ekohl, line 266: The indenting looks off here

girijaasoni: No, it's fine I had cross checked

lib/fog/vsphere/compute.rb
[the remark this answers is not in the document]
girijaasoni: We technically don't need to re-enable it when inline but since we've been using that enable - disable format throughout, I have modified it

lib/fog/vsphere/requests/compute/create_vm.rb
chris1984: This looks off if I look at the file, could it be your editor? https://github.com/fog/fog-vsphere/blob/9b71befc89ddf26021daadfdf86f7c1b3c2be6d1/lib/fog/vsphere/requests/compute/create_vm.rb#L266

girijaasoni: Updated with a new indentation. Either way, was accepted by rubocop and the general editor format

chris1984:
Looks fine to me @ekohl any other thoughts before I do a final round of testing.

.rubocop.yml
```diff
@@ -42,3 +42,7 @@ Style/RescueModifier:
 
 Metrics/MethodLength:
   Max: 45
+
+Style/MutableConstant:
```
ekohl, line 46: I'd be hesitant to disable this. Why is it needed?

girijaasoni: used the freeze method for the string constants

chris1984:
@ekohl any other comments or is this good to go, tested it and it works with the latest commit and subhash did as well.

ekohl:
@chris1984 I still see open conversations when I look on the changed files tab. I'd expect those to either be fixed or closed with an explanation why it won't be fixed.

lib/fog/vsphere/requests/compute/get_vm_first_nvme_controller.rb
girijaasoni: We are not doing that for the other(scsi and sata) controllers, I dont think it's necessary to do that as it's the mock class.

girijaasoni:
> @chris1984 I still see open conversations when I look on the changed files tab. I'd expect those to either be fixed or closed with an explanation why it won't be fixed.

Thanks @ekohl , updated them.

chris1984:
Looks fine now, just one concern:

I tested this again and I am getting Failed to create a compute VMware8 (VMware) instance [ron-doria.satellite.lab.eng.rdu2.redhat.com](http://ron-doria.satellite.lab.eng.rdu2.redhat.com/): InvalidController: The device '1' is referring to a nonexisting controller '1,000'.


Since it keeps happening between my env and qe's I wonder if customers will start to hit this too. Is there a way we can get that controller id for all envs since you make a fix and it works on mine, but errors's on qe's vmware then you fix and it works on theirs but not mine. I am worried customer will hit that same error

chris1984 (approved):
ACK, tested it on 6/7/8 and it works fine as well as on QE's VMware 8:

Screenshot 2024-07-11 at 08-34-20 vSphere - cheri-ottosen satellite lab eng rdu2 redhat com - Summary
Screenshot 2024-07-11 at 08-28-59 vSphere - annie-brigman satellite lab eng rdu2 redhat com - Summary
Screenshot 2024-07-11 at 08-40-22 vSphere - pedro-hunkele satellite lab eng rdu2 redhat com - Summary
Screenshot 2024-07-11 at 08-41-44 vSphere - teri-hargus satellite lab eng rdu2 redhat com - Summary

PR #302, opened 2024-07-15
Add Foreman gem push workflow

chris1984:


.github/workflows/gem_push.yml
```diff
@@ -0,0 +1,16 @@
...
+    # Pattern matched against refs/tags
+    tags:
+      - '**'
+
+jobs:
+  release:
+    name: Release gem
+    uses: theforeman/actions/.github/workflows/release-gem.yml@v0
+    with:
+      allowed_owner: chris1984
```
ekohl, line 14: This can't work because it's the GH owner so this only works on your forks. You can see it never ran: https://github.com/fog/fog-vsphere/actions/workflows/gem_push.yml

```suggestion
      allowed_owner: fog
```

PR #303, opened 2024-07-15
Add Gemfile and gemspec to gem build files so rpm can be built.

chris1984:


ekohl:
I only see this now, but you do not need them for RPMs to build. Instead what should have been done is to drop them from the `%files` section in the spec. You're effectively reverting 4a7f4e7f96b41190a5b6733112a84f8af8aabb6d.

PR #305, opened 2024-07-18
Secure Boot & Virtual TPM

nofaralfasi:
Add support for Secure Boot & Virtual TPM.


chris1984 (commented):
Code looks fine, when you get the Foreman one made and it's ready to go, i'll test and ack based on testing.

lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -22,7 +22,7 @@ def create_vm(attributes = {})
           vm_cfg[:cpuHotAddEnabled] = attributes[:cpuHotAddEnabled] if attributes.key?(:cpuHotAddEnabled)
           vm_cfg[:memoryHotAddEnabled] = attributes[:memoryHotAddEnabled] if attributes.key?(:memoryHotAddEnabled)
           vm_cfg[:firmware] = attributes[:firmware] if attributes.key?(:firmware)
-          vm_cfg[:bootOptions] = boot_options(attributes, vm_cfg) if attributes.key?(:boot_order) || attributes.key?(:boot_retry)
+          vm_cfg[:bootOptions] = boot_options(attributes, vm_cfg)
```
chris1984, line 25: What was the reason we are dropping the if statement here for the boot order?

chris1984:
Looks fine just 1 question, starting to test

lib/fog/vsphere/requests/compute/create_vm.rb
nofaralfasi: The `boot_options` method will not be called if `boot_order` or `boot_retry` are not present, but we need this method to be called for Secure Boot configuration. Also,  `boot_order` is always passed by default, so the condition is unnecessary.

chris1984 (commented):
Code looks fine, we are getting a rubocop error:
`Error: Metrics/ClassLength: Class has too many lines. [290/284]`

Feel free to disable the cop for that method and repush and I think we can merge this, pending the question I sent you on Slack.

PR #307, opened 2024-12-10
Fix warning on nil comparison

ekohl:
This uses the more efficient /regex/.match?(value) syntax that also doesn't raise any warning if id is nil.

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -347,7 +347,7 @@ module Fog
         end
 
         def is_uuid?(id)
-          !(id =~ /[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}/).nil?
+          /[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}/.match?(id)
         end
       end
 


chris1984 (approved):

ekohl:
@chris1984 are you sure the change didn't cause those test failures?

chris1984:
Blah, I did not mean to hit merge, was adding that I would start testing it with the quick responses and my browser crashed. Let me revert and test it, if it's not related we can remerge.

nofaralfasi:
I believe this change is causing errors in Foreman's test suite.

ekohl:
> I believe this change is causing errors in Foreman's test suite.

Can you elaborate?

nofaralfasi:
This change introduces the following error, when running the tests on `foreman/test/models/compute_resources/vmware_test.rb`:

```
Error:
Foreman::Model::VmwareTest::#clone_vm#test_0004_passes customspec when user_data is a valid yaml hash:
TypeError: no implicit conversion of Integer into String
    app/models/compute_resources/foreman/model/vmware.rb:605:in `clone_vm'
    test/models/compute_resources/vmware_test.rb:715:in `block (2 levels) in <class:VmwareTest>'

Error:
Foreman::Model::VmwareTest::#clone_vm#test_0003_ignores customspec when user_data is nil:
TypeError: no implicit conversion of Integer into String
    app/models/compute_resources/foreman/model/vmware.rb:605:in `clone_vm'
    test/models/compute_resources/vmware_test.rb:706:in `block (2 levels) in <class:VmwareTest>'
```

We are encountering this error when client.servers.get is called on an Integer ([here](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L605)).

ekohl:
That could be related, but previously that also generated a warning I think. A better check is probably `is_a?(String) && ...`.

ekohl:
Expanding on that: it was a bug before anyway on modern Ruby anyway:
```console
$ pry
[1] pry(main)> 5 =~ /a/
NoMethodError: undefined method `=~' for an instance of Integer (NoMethodError)
from (pry):1:in `__pry__'
```

IIRC Ruby 3.0 already displayed a deprecation warning and this was trying to address it. I'll submit a PR.

ekohl:
https://github.com/fog/fog-vsphere/pull/310

PR #308, opened 2025-01-06
Revert "Fix warning on nil comparison"

chris1984:
Reverts fog/fog-vsphere#307


ekohl:
Looks like this fails in the same way so my chance probably didn't cause it.

PR #313, opened 2025-06-19
Use each_with_index instead of calculating indices

ekohl:


ekohl:
https://github.com/fog/fog-vsphere/pull/314 should fix the tests.

stejskalleos (approved):

PR #316, opened 2026-02-07
Add request to delete iso file

sbernhard:


lib/fog/vsphere/requests/compute/destroy_iso.rb
```diff
@@ -0,0 +1,45 @@
+module Fog
+  module Vsphere
+    class Compute
+      class Real
+        def destroy_iso_check_options(options)
```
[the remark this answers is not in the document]
stejskalleos: Oh, I guess it's for `destroy_iso`.
Then I suggest moving it to private methods and simply renaming it `check_options'!`

stejskalleos (changes requested):

fog-vsphere.gemspec
```diff
@@ -25,8 +25,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'fog-core'
   spec.add_runtime_dependency 'rbvmomi2', '~> 3.0'
 
-  spec.add_dependency "base64"
   spec.add_dependency "ostruct"
+  spec.add_dependency "base64"
```
nadjaheitmann: Is this really necessary? I mean moving the line around although it was added in the commit before..

lib/fog/vsphere/requests/compute/destroy_iso.rb
```diff
@@ -0,0 +1,65 @@
...
+          remote_ds  = "[#{options['datastore']}] #{remote_rel}"
+
+          task = connection.serviceContent.fileManager.DeleteDatastoreFile_Task(
+            name: remote_ds,
+            datacenter: datacenter
+          )
+          task.wait_for_completion
+
+          # Return true if it is gone
+          !datastore.exists?(remote_rel)
```
nadjaheitmann: If something prevents the the task from completion, what happens? Is there any fallback or will the task stay there in a pending state?

stejskalleos (approved):
🍏 Code LGTM,
I haven't tested it personally, but we have ACK from @nadjaheitmann from other PR (https://github.com/theforeman/foreman_bootdisk/pull/181) as well, so let's get this in.

